PR #32, opened 2013-08-04
Sets content-type to text/xml

HadesArchitect:
Content-type should be a text/xml type, not plain text


ostrolucky:
For anybody running into this: If you use FOS restbundle, you probably need to adjust format listener configuration https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Resources/doc/format_listener.rst

PR #53, opened 2014-02-12
Added filter_var instead of direct checking

Warxcell:
I don't really like annotation limitation for routes marked for sitemap. Now you can use XML too:

<route id="route" pattern="/">
        <default key="_controller">AcmeDemoBundle:Default:home</default>
        <option key="sitemap">true</option>
</route>


Koc:
@Warxcell @nicolas-bastien @BenoitLeveque can this PR can be closed because #81 was merged?

nicolas-bastien:
@Koc ok thanks

PR #64, opened 2014-10-08
feat(base_url): fallback to symonfy router context instead of hardcoded localhost

cfuerst:
Since Symfony suggests this (http://symfony.com/doc/current/cookbook/console/sending_emails.html) to use base paths outside request context, i would recommend to use the router context for fallback to localhost.

You would be able to use your parameters.yml config without specifying it in config or in the command. The change is fully BC, since router context falls back to http and localhost anyway.

```
# app/config/parameters.yml
parameters:
    router.request_context.host: mysite.ch
    router.request_context.scheme: https
```

Would do the job to start the dump without any config in presta or specifying urls in the command.

Chers


cfuerst:
thnx

PR #81, opened 2015-05-01
Improve #53 - Added filter_var instead of direct checking

partikus:
It's allow to use sitemap option in the xml file. 
Please let me know if any other test cases are needed. 
@alain-flaus code review pls

### DIFF
--- a/Tests/EventListener/RouteAnnotationEventListenerTest.php
+++ b/Tests/EventListener/RouteAnnotationEventListenerTest.php
@@ -34,7 +34,7 @@ class RouteAnnotationEventListenerTest extends \PHPUnit_Framework_TestCase
      */
     public function testInvalidSitemapFalse()
     {
-        $this->setExpectedException('\InvalidArgumentException');
+        $this->setExpectedException('InvalidArgumentException');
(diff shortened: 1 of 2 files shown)


partikus:
@alain-flaus ping

partikus:
Please take a look on that, thx 
 @esion @alain-flaus @kostiklv @Koc

Tests/EventListener/RouteAnnotationEventListenerTest.php
```diff
@@ -34,7 +34,7 @@ public function testNoAnnotation()
      */
     public function testInvalidSitemapFalse()
     {
-        $this->setExpectedException('InvalidArgumentException');
+        $this->setExpectedException('\InvalidArgumentException');
```
Koc: IMHO there is no sense for backslash here

partikus: I propose to unify all usages of ```InvalidArgumentException``` in the file. 
We can use ```InvalidArgumentException``` for each usage(without backslash). 
@Koc What do you think?

Koc: Better remove backslashes when using class name as string. No sense for it.

partikus:
@Koc done and fixed

partikus:
ping @Koc @alain-flaus

PR #83, opened 2015-05-16
Fix deprecated routing config options

xaben:
Fixes deprecation messages for Symfony 2.7.

This bumps the minimum symfony version to 2.2, as pattern was replaced by path since 2.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ php:
     - 5.5
 
 env:
-    - SYMFONY_VERSION=2.1.*
     - SYMFONY_VERSION=2.2.*
     - SYMFONY_VERSION=2.3.*
     - SYMFONY_VERSION=2.4.*
     - SYMFONY_VERSION=2.5.*
+    - SYMFONY_VERSION=2.6.*
     - SYMFONY_VERSION=dev-master
 
 before_script:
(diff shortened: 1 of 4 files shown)


xaben:
@BenoitLeveque Can you please review this PR? Thanks!

yann-eugone:
Not so easy.
You just can't remove 2.1 support on the master branch.
We must choose wisely the way to handle the 2.7 deprecation...

I recommand creating a dedicated branch for the 2.1 to 2.6 version.

xaben:
Well considering Symfony 2.1 support ended in November 2013 and you plan to continue using it, then yeah tagging a release that supports 2.1 or splitting into a different branch is the way to go.

Koc:
+1 for dropping Symfony 2.1 support

DZunke:
+1 bumping minimum version. at a minimum the master should be set to the actual LTS-Version of Symfony2. So the best would be 2.3 (2017) or 2.7 (2019)

http://symfony.com/roadmap#checker

chteuchteu:
+1 for dropping 2.1 support. Plus,

 > The "pattern" option is deprecated since version 2.2 and will be removed in 3.0. Use the "path" option in the route definition instead.

Koc:
@yann-eugone this PR replaced by #93 and can be closed

PR #60, opened 2014-07-18
Added PHP 5.6 and HHVM to travis.yml

Nyholm:
I did also fix the build matrix to reduce the number of builds.

PR #93, opened 2015-11-11
Improve Travis & Composer config

WouterJ:
**Travis**

 * Use the faster docker based Travis environment + the caching feature it provides
 * Test against all supported Symfony and PHP versions
 * Instead of testing all Symfony versions against all PHP versions, test all PHP versions against one Symfony version (latest stable) and test all Symfony versions against one PHP version (latest stable).
 * Instead of running phpcs every job, have a special job for phpcs (even better would be to use StyleCI instead of Travis)

**Composer**

 * Do not claim Symfony >=2.1.0 support as you don't know if this bundle supports Symfony 10.
 * Use PSR-4 instead of the deprecated target-path and PSR-0 autoloading logic

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,29 +4,55 @@ php:
     - 5.3
     - 5.4
     - 5.5
-
-env:
-    - SYMFONY_VERSION=2.1.*
-    - SYMFONY_VERSION=2.2.*
-    - SYMFONY_VERSION=2.3.*
-    - SYMFONY_VERSION=2.4.*
-    - SYMFONY_VERSION=2.5.*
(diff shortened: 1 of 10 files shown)


WouterJ:
I also dropped Symfony 2.1 compatibility. It's hard to support 2.2 and 3.0 and I believe you shouldn't worry about supporting versions that reached end of life (e.g. Symfony <2.3, 2.4, 2.5 and 2.6).

Koc:
closes #60

WouterJ:
Tests pass, so it seems like you support Symfony 3 after this PR is merged (and PHP 7). :tada:

Koc:
:clap:

PR #99, opened 2016-02-24
Preserve Previous Exception in getRouteUri

lemoinem:
This usually helps a lot with debugging by preserving the stack trace. There is no reason to break the stack trace.


yann-eugone:
Solved in #104

PR #117, opened 2016-06-13
Make Sitemap Event Listeners private

bocharsky-bw:
I think we need to mark Sitemap Event Listeners as private to get a minor performance boost.


bocharsky-bw:
Looks like it doesn't work as a private

yann-eugone:
Symfony's event listeners have to be public, as these are lazy loaded

PR #120, opened 2016-06-13
Add example of YAML configuration

bocharsky-bw:


yann-eugone:
I think that we also could add support for XML and more complicated value than `true` or `false`.
I will push as comment some examples

yann-eugone:
```yml
yml_homepage:
    path:     /
    defaults: { _controller: AppBundle:Default:index }
    options:
        sitemap: true
```

```yml
yml_homepage:
    path:     /
    defaults: { _controller: AppBundle:Default:index }
    options:
        sitemap:
            section: yml
```

```xml
<route id="xml_homepage" path="/">
    <default key="_controller">AppBundle:Default:index</default>
    <option key="sitemap">true</option>
</route>
```

```xml
<route id="xml_homepage" path="/">
    <default key="_controller">AppBundle:Default:index</default>
    <option key="sitemap">
        {"section":"xml"}
    </option>
</route>
```

bocharsky-bw:
@yann-eugone Hey, thanks for this examples! I've added simple YAML and XML examples of route configuration here in this PR. Maybe expanded configuration examples better move to the separate docs page instead of README? What do you think?

yann-eugone:
@bocharsky-bw yes, I agree with this. i will do it later, can you open an issue ?

bocharsky-bw:
@yann-eugone I did!

PR #123, opened 2016-06-15
Change default URL priority to the 0.5

bocharsky-bw:
The default page priority should be 0.5, check docs here http://www.sitemaps.org/protocol.html#xmlTagDefinitions .

WARNING: It entails possible BC breaks.

### DIFF
--- a/Sitemap/Url/UrlConcrete.php
+++ b/Sitemap/Url/UrlConcrete.php
@@ -42,7 +42,7 @@ class UrlConcrete implements Url
      * @param string $changefreq
      * @param float $priority
      */
-    public function __construct($loc, \DateTime $lastmod = null, $changefreq = null, $priority = null)
+    public function __construct($loc, \DateTime $lastmod = null, $changefreq = null, $priority = 0.5)
     {
         $this->setLoc($loc);
         $this->setLastmod($lastmod);


yann-eugone:
Doing this meens that priority will never be nullable anymore.
Seems pretty dangerous, no ?

bocharsky-bw:
Actually, for now, with simple configuration with annotation like `"options": {"sitemap": true}` we have default value `1`. Here's output:
```xml
<url>
<loc>http://example.com/</loc>
<lastmod>2016-06-15T19:26:57+03:00</lastmod>
<changefreq>daily</changefreq>
<priority>1.0</priority>
</url>
```

bocharsky-bw:
@yann-eugone Sorry, I confused you a bit! Proper changes should be done in annotation listener, I did it in https://github.com/prestaconcept/PrestaSitemapBundle/pull/123/commits/9029e591f3c0fac808543425cd7902cac7df6b22 .

bocharsky-bw:
But I think it again... IMO, it's strange to have this default values for `changefreq` (daily) and `priority` (1) for annotations. I think we need to set `null` for them by default, as already did in the `UrlConcrete`.

yann-eugone:
I noticed that the behavior of urls registered via annotations and urls registered via custom ways do not have the same output, because of the defaults values, handled as default by the RoutingAnnotationListener.

I think we have 2 possibilities : 
- no default values for any of them
- default values for all routes registered (meen that added routes must be post-processed with default values)

bocharsky-bw:
I think we should skip default values for options which are not required. Or we can add ability to set default values in bundle configuration.

yann-eugone:
I already added the defaults as [configuraton values](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/DependencyInjection/Configuration.php#L51).

But this is not applied to the route that are registered with any custom loader.
The piece of code that handle the "default" is [here](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/EventListener/RouteAnnotationEventListener.php#L153).

That is why I proposed to handle defaults as "post process", for example in the [`addUrl` method](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/AbstractGenerator.php#L63).

yann-eugone:
@bocharsky-bw I created #129 , the goal is to let the defaults, being added in the addUrl method.
That way, from anywhere you add an url to the sitemap, you are sure that default values will be added to your URL.

What do you think ?

bocharsky-bw:
👍 It makes sense. I think we can close current #123 in favor of yours.

yann-eugone:
Why ?
If it is recommanded that the default value of priority is 0.5 as default, we better follow these recommandations, no ?

bocharsky-bw:
Ah, yes, you're right! But we also have a BC breaks here

yann-eugone:
You mean, changing the default value for priority ?

bocharsky-bw:
Yes, exactly. I think many devs simply use default values.

yann-eugone:
So what, we just leave the priority as 1 as default ?

bocharsky-bw:
I think we can keep this PR for the next major release and mark it with a proper tag (i.e. `break_bc` or `waiting_major_release`). How does it sound?

yann-eugone:
ok fine, i added the label

kiler129:
A lot of things changed since 2016, I think this PR is not needed, since search engines ignore priority:
 - https://www.seroundtable.com/google-sitemap-priority-field-a-bag-of-noise-23645.html
 - https://www.searchenginejournal.com/xml-sitemap-best-practices/237649/#close

bocharsky-bw:
Then it would not even be a BC change, but I think matching docs defaults would be good anyway

PR #134, opened 2016-09-29
Deprecated SitemapListenerInterface

### LINKED ISSUE #131: SitemapListenerInterface is redurant

yann-eugone:
in favor of Symfony standard event listener/subscriber registering

fixes #131

### DIFF
--- a/EventListener/RouteAnnotationEventListener.php
+++ b/EventListener/RouteAnnotationEventListener.php
@@ -12,8 +12,8 @@
 namespace Presta\SitemapBundle\EventListener;
 
 use Presta\SitemapBundle\Event\SitemapPopulateEvent;
-use Presta\SitemapBundle\Service\SitemapListenerInterface;
 use Presta\SitemapBundle\Sitemap\Url\UrlConcrete;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\Route;
@@ -21,8 +21,6 @@ use Symfony\Component\Routing\RouteCollection;
 use Symfony\Component\Routing\RouterInterface;
 
 /**
- * Class RouteAnnotationEventListener
- *
  * this listener allows you to use annotations to include routes in the Sitemap, just like
  * https://github.com/dreipunktnull/DpnXmlSitemapBundle
  *
@@ -36,9 +34,8 @@ use Symfony\Component\Routing\RouterInterface;
  * Route("/", name="homepage", options={"sitemap" = true })
  *
  * @author Tony Piper (tpiper@tpiper.com)
- * @license see prestaConcept license
  */
-class RouteAnnotationEventListener implements SitemapListenerInterface
+class RouteAnnotationEventListener implements EventSubscriberInterface
 {
     /**
      * @var RouterInterface
@@ -56,7 +53,17 @@ class RouteAnnotationEventListener implements SitemapListenerInterface
     /**
      * @inheritdoc
      */
-    public function populateSitemap(SitemapPopulateEvent $event)
+    public static function getSubscribedEvents()
+    {
+        return [
+            SitemapPopulateEvent::ON_SITEMAP_POPULATE => ['registerRouteAnnotation', 0],
+        ];
+    }
+
+    /**
+     * @param SitemapPopulateEvent $event
+     */
+    public function registerRouteAnnotation(SitemapPopulateEvent $event)
     {
         $section = $event->getSection();
 
--- a/EventListener/RouteAnnotationEventListener.php
+++ b/EventListener/RouteAnnotationEventListener.php
@@ -55,9 +55,9 @@ class RouteAnnotationEventListener implements EventSubscriberInterface
      */
     public static function getSubscribedEvents()
     {
-        return [
+        return array(
             SitemapPopulateEvent::ON_SITEMAP_POPULATE => ['registerRouteAnnotation', 0],
-        ];
+        );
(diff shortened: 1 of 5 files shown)


yann-eugone:
@Koc what do you think ?

EventListener/RouteAnnotationEventListener.php
```diff
@@ -56,7 +53,17 @@ public function __construct(RouterInterface $router)
     /**
      * @inheritdoc
      */
-    public function populateSitemap(SitemapPopulateEvent $event)
+    public static function getSubscribedEvents()
+    {
+        return [
```
Koc: Breaks 5.3 PHP which is already supported

Service/SitemapListenerInterface.php
```diff
@@ -17,6 +17,10 @@
  * Inteface for sitemap event listeners
  *
  * @author Konstantin Tjuterev <kostik.lv@gmail.com>
+ *
+ * @deprecated This class has been deprecated in favor of Symfony standard event listener and subscriber.
```
Koc: Interface

Resources/doc/5-Usage-Event_Listener.md
```diff
@@ -1,75 +1,110 @@
-# Usage Sitemap Event Listeners
+# Sitemap Events Usage
 
-You can also register your sitemap event listeners by creating service classes implementing
-`Presta\SitemapBundle\Service\SitemapListenerInterface` and tagging these services with `presta.sitemap.listener`
-tag in your `Resources/config/services.xml`. This way the services will be lazy-loaded by Symfony's event dispatcher, only when the event is dispatched:
+You can also register event listeners (or subscribers) to populate your sitemap(s).
```
Koc: New example is really better than old :+1: .

Resources/doc/5-Usage-Event_Listener.md
```diff
@@ -1,75 +1,110 @@
...
+    /**
+     * @var EntityManager
+     */
+    private $manager;
+
+    /**
+     * @param RouterInterface $router
+     * @param EntityManager   $manager
+     */
+    public function __construct(RouterInterface $router, EntityManager $manager)
```
Koc: it is better use `UrlGeneratorInterface` and `EntityManagerInterface`. Also in all examples entity manager named like `$em`

Resources/doc/5-Usage-Event_Listener.md
```diff
@@ -1,75 +1,110 @@
...
+            SitemapPopulateEvent::ON_SITEMAP_POPULATE => 'registerBlogPostsPages',
+        ];
+    }
+
+    /**
+     * @param SitemapPopulateEvent $event
+     */
+    public function registerBlogPostsPages(SitemapPopulateEvent $event)
+    {
+        $posts = $this->manager->getRepository('AppBundle:BlogPost')->findAll();
```
Koc: Can we add note for big results set that better use `iterate` instead of loading all result set http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/batch-processing.html#iterating-results ? Or maybe hydrate as array.

Resources/doc/5-Usage-Event_Listener.md
yann-eugone: We can even use the ObjectManager interface.
I don't agree with $em var name, it's too short and is not compliant with psr code style.

Resources/doc/5-Usage-Event_Listener.md
yann-eugone: We can leave a note  (or disclaimer) that using findAll method may not be the best solution for large amount of data. 
But this is not up to us finding the good way to fetch data.

Koc: note will be enough

Resources/doc/5-Usage-Event_Listener.md
```diff
@@ -1,75 +1,110 @@
...
-<services>
-    <service id="my.sitemap.listener" class="%acme_demo.sitemap.listener.class%">
-        <tag name="presta.sitemap.listener" />
-        <argument type="service" id="router"/>
-    </service>
-</services>
-```
-
-or in yaml:
+**note :** we choose an `event subscriber` as example, but you can also do it with an `event listener`.
```
lemoinem: Since the subscriber is supporting a single event why prefer a subscriber over a listener?

yann-eugone: Symfony's event subscribers are supporting as many event as you want, in fact the only benefit of using a subscriber instead of a listener, is that you will be allowed to use event constants.

yann-eugone:
@Koc @lemoinem something more ?

Resources/doc/5-Usage-Event_Listener.md
```diff
@@ -78,6 +78,9 @@ class SitemapBlogPostSubscriber implements EventSubscriberInterface
 }
 ```
 
+**note :** you may not use this snippet as is. With large dataset, `findAll` is not a good idead.
```
lemoinem: Maybe "should" instead of "may"? It's nitpicking, really, but "may" sounds way too authoritative for something we just want to mention... ??

Resources/doc/5-Usage-Event_Listener.md
Koc: not only. Subscriber also knows on what events with what priority it should be injected. It is useful. Also it is useful when you reuse thirdparty subscribers - just tag them.

Koc:
@yann-eugone thank you for your work.

PR #135, opened 2016-09-29
Fixed Travis CI build

yann-eugone:


yann-eugone:
We decided to drop support of old php versions, in future releases  (future when we dropped it)
Reminder : PHP did drop support for versions under 5.6 (see http://php.net/supported-versions.php)

Despite this, I saw that the composer.json was not  uodated...

Koc:
It is better to drop old php versions when bumping budle version to 2.0 for respecting semver

PR #139, opened 2016-11-29
tipo(doc): Added full command name

ErikSaunier:


yann-eugone:
thank you for the catch

PR #143, opened 2017-08-03
Update GoogleImageUrlDecorator.php

Warxcell:
Bug fix - exception class is not found.

PR #144, opened 2017-08-03
Update and rename GoogleImageException.php to GoogleImageUrlDecorator…

Warxcell:
….php

Maybe this is the correct name for the exception? It's used in: https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Sitemap/Url/GoogleImageUrlDecorator.php#L35


yann-eugone:
@Warxcell thank you for reporting this. This patch should be included in your first pull request : #143 
I'm closing this one so.

yann-eugone:
@Warxcell sorry i did not read carrefully, but still, #143 seems more relevant to me

PR #148, opened 2017-08-07
Bugfix - Exception class not found

Warxcell:
Using `Presta\SitemapBundle\Exception\GoogleImageException` instead of non-existing
`Presta\SitemapBundle\Exception\GoogleImageUrlDecorator `


yann-eugone:
Nice, thank you :)

PR #150, opened 2017-11-22
Version 2.0

yann-eugone:
What's in :

- [x] PHP7 compliance
- [x] Symfony 3.4 & 4.0 compliance
- [x] Deprecations removal
- [ ] Usage of Symfony Cache


Koc:
Let's bump to 7.1, please. <3 nullable typehits/return types

Koc:
http://php.net/supported-versions.php
Active support of PHP 7.0 ends in 10 days.

Also what about removing some useless phpdocs? Typehints enough in most cases.

yann-eugone:
Agreed.

But Symfony PHP requirement la still on 7.0 (and they are not about to change it : https://medium.com/@fabpot/symfony-4-an-update-on-flex-d12716ff8b92 )

It is an open debate : should we ask for an higher version of PHP than Symfony do?

Koc:
You mixed up Flex requirements and Symfony 4.0 https://github.com/symfony/symfony/blob/4.0/composer.json#L19

4.0 already requires 7.1

yann-eugone:
Yeah, you are also right.
But still, Symfony `3.4` PHP requirement is `>=7.0.8`

Koc:
Anybody who using Symfony 3.4 can stay on current 1.x branch. Let's moving to the future.

Koc:
To be honest, Symfony 3.4 requirement is: `^5.5.9|>=7.0.8`, so should we support php 5.5.9 in 2.0 version of bundle?)

yann-eugone:
The more we talk about it, the less I think this major version will be usefull.
If this is all about having scalar typehints, this is not worth it...
I refuse to maintain 2 branches just for fun.

yann-eugone:
@Koc I created #151 with support for Symfony 3.4 and 4.0.
As you can see, current 1.5 version is not yet compatible : deprecation warnings most of the time.
So the main changes for this PR we are discussing of is scalar typehint...

Don't you think this is enough ?

Koc:
Commands and controllers as services also good step. But I agree, that supporting both branches is harder than one.

yann-eugone:
closing in favor of #151

PR #152, opened 2017-11-27
Enhances PhpDoc

yann-eugone:
from #150

### DIFF
--- a/Service/Dumper.php
+++ b/Service/Dumper.php
@@ -12,10 +12,11 @@
 namespace Presta\SitemapBundle\Service;
 
 use Presta\SitemapBundle\DependencyInjection\Configuration;
+use Presta\SitemapBundle\Sitemap\DumpingUrlset;
+use Presta\SitemapBundle\Sitemap\Urlset;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
-use Presta\SitemapBundle\Sitemap\DumpingUrlset;
 
 /**
  * Service for dumping sitemaps into static files
@@ -27,14 +28,12 @@ class Dumper extends AbstractGenerator implements DumperInterface
 {
     /**
      * Path to folder where temporary files will be created
-     *
      * @var string
      */
     protected $tmpFolder;
 
     /**
      * Base URL where dumped sitemap files can be accessed (we can't guess that from console)
-     *
      * @var string
      */
     protected $baseUrl;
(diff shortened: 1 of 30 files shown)


Service/Generator.php
```diff
@@ -45,12 +46,18 @@ class Generator extends AbstractGenerator implements GeneratorInterface
...
-     * @param integer|null             $cacheTtl
-     * @param integer|null             $itemsBySet
+     * @param int|null                 $cacheTtl
+     * @param int|null                 $itemsBySet
      */
-    public function __construct(EventDispatcherInterface $dispatcher, RouterInterface $router, Cache $cache = null, $cacheTtl = null, $itemsBySet = null)
-    {
+    public function __construct(
+        EventDispatcherInterface $dispatcher,
+        RouterInterface $router,
```
Koc: Can we use `UrlGeneratorInterface` here (interface segregation)?

Sitemap/Url/UrlConcrete.php
```diff
@@ -173,11 +209,10 @@ public function toXml()
...
 
     /**
-     * basic url has no namespace. see decorated urls
-     * @return array
+     * @inheritdoc
      */
     public function getCustomNamespaces()
     {
-        return array();
+        return array(); // basic url has no namespace. see decorated urls
```
Koc: Some arrays converted to short arrays, some not 😕

Service/Dumper.php
```diff
@@ -89,14 +89,17 @@ public function dump($targetDir, $host, $section = null, array $options = array(
         }
 
         foreach ($this->urlsets as $urlset) {
-            $urlset->save($this->tmpFolder, $options['gzip']);
+            if ($urlset instanceof DumpingUrlset) {
```
Koc: Should we throw an exception if `Dumper` got something else than `DumpingUrlset`?

Sitemap/Url/GoogleNewsUrlDecorator.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Presta\SitemapBundle\Sitemap\Url;
 
+use DateTime;
```
Koc: Symfony usually does not import classes from global namespace

yann-eugone: Since it not in any pratice guide, I assume this is matter of preference.

Service/Dumper.php
yann-eugone: We may, but this is not the purpose of this PR.

Sitemap/Url/UrlConcrete.php
yann-eugone: I tried not to write short syntax arrays, where did you find some in the introduced code ?

We may do an other PR with array short syntax uniformisation

Service/Generator.php
yann-eugone: Agreed.

PR #157, opened 2018-01-29
Allow lastmod to be nullable

maff:
Currently the `lastmod` property is not nullable because even if setting the default to `null`, it will be created as `new \DateTime(null)` which resolves to now. This adds a check if the default is actually `null` + tests for the defaults handling.


yann-eugone:
Seems OK to me. Keeping it open for a few days for the team to have a look.

Thank you BTW

yann-eugone:
Thank you @maff for this catch

Issue #146, opened 2017-08-03
Prepare next major version


yann-eugone:
**WIP**

**Versions**

Compatibility matrix became horrible. We would like to support less versions, but supporting it better.

Objective : 
- Symfony : >= 3.4
- PHP >= 7.0

**Related issues**

- [Symfony4 Support](https://github.com/prestaconcept/PrestaSitemapBundle/pull/149)
- [Rely on Symfony Cache](https://github.com/prestaconcept/PrestaSitemapBundle/issues/142)

yann-eugone:
Closing this.
V2 is out, but we didn't had time to work on cache BC.

Issue #142, opened 2017-07-10
Use Symfony Cache Component instead of DoctrineCache


kopeckyales:
Since Symfony 3.1, Symfony provides cache component out of the box. It would be nice to drop dependency on DoctrineCache and use default symfony cache component for caching sitemaps.

Any plans to implement this?

unexge:
i am using this dirty hack while waiting #160 to merge.

```yaml
# config/packages/presta_sitemap.yaml
services:
    app.presta_sitemap.cache:
        class: Symfony\Component\Cache\DoctrineProvider
        arguments:
            - "@cache.app" # using app cache pool my cause collisions, you can use custom cache pool

    app.presta_sitemap.generator:
        class: Presta\SitemapBundle\Service\Generator
        autowire: true
        arguments:
            - "" # autowired
            - "" # autowired
            - "@app.presta_sitemap.cache"
            - "%presta_sitemap.timetolive%"
            - "%presta_sitemap.items_by_set%"
        calls:
            - method: setDefaults
              arguments:
                - '%presta_sitemap.defaults%'

    # if you want to autowire Generator service
    Presta\SitemapBundle\Service\Generator: '@app.presta_sitemap.generator'

presta_sitemap:
    generator: app.presta_sitemap.generator
    # ...
```

yann-eugone:
As proposed in #241 : it was decided to drop cache support in favor of sitemap dump.

As a result, this issue is closed

PR #160, opened 2018-02-18
Drop DoctrineCacheBundle in favor of Symfony Cache component

enekochan:
This PR refers to https://github.com/prestaconcept/PrestaSitemapBundle/issues/142 and https://github.com/prestaconcept/PrestaSitemapBundle/issues/146

I've used a compiler pass to inject the cache service because its name is used in the configuration and it's the only way I could think off. Had tried using expressions in the Generator service configuration but couldn't make it work.

Would be nice if people with real big sites and various cache systems could test this :)


yann-eugone:
We must introduce compatibility with Symfony cache system. You're right. But doing the things like this will not help people upgrading.

We must think about deprecating the usage of doctrine cache, triggering appropriate error if used.

Could you consider put the whole cache implementation in a dedicated class that accept both doctrine cache and symfony cache, (wrapping the doctrine implementation with an adapter for instance) and trigger deprecation notices ?

A note should also be added to tell people how it was before, so they see what they have to do to upgrade.

ostrolucky:
@enekochan why did you decide to use symfony cache interfaces, instead of generic PSR ones?

enekochan:
@ostrolucky Because the Symfony Cache interfaces use internally the PSR ones and I was intending to add the use of Symfony Cache component so I used it as is. But may be there is a better way to do this that I don't know.

enekochan:
@yann-eugone I've been out for some time. I'll try to think of a solution that allows both cache methods and add the deprecation notices. Hope I can upload something soon.

ostrolucky:
For compatibility, some instanceof checks are ok. Doctrine one could invoke deprecation.

For PSR, probably all you need is to just depend on `\Psr\Cache\CacheItemPoolInterface`

yann-eugone:
Internally, wa can rely on a PSR cache interface. We will need to detect what kind of service will configure for us.

If the service is a doctrine cache, we will wrap it up with a https://github.com/php-cache/doctrine-adapter implementation.
Doing this should trigger a deprecation to help developers to update their code (even is the cache PSR standard was not a clear victory vor everyone).

The documentation should be updated according to the new practices / directories.

fnowacki:
@yann-eugone what about this? What's status?

yann-eugone:
The same as 6 months ago. I didn't had time to work on this (actually I pretty much never use that feature : dump is my way to go).

And I'm still not open of just dropping some library support in favor of another one.

TLDR : We still need some work to make upgrading from the "old" way to the "new" one.

yann-eugone:
As proposed in #241 : it was decided to drop cache support in favor of sitemap dump.

As a result, this PR is closed

PR #167, opened 2018-08-15
Set default location to public

johnhout:
Since symfony 4, public folder is default.

### DIFF
--- a/Command/DumpSitemapsCommand.php
+++ b/Command/DumpSitemapsCommand.php
@@ -55,7 +55,7 @@ class DumpSitemapsCommand extends ContainerAwareCommand
                 'target',
                 InputArgument::OPTIONAL,
                 'Location where to dump sitemaps. Generated urls will not be related to this folder.',
-                'web'
+                'public'
             );
     }
 


yann-eugone:
True thing. But this will introduce a BC break if we do it like this.
Maybe we can try to detect whether or not the default value should be `web` or `public` at runtime.

Maybe using [version_compare](http://php.net/manual/fr/function.version-compare.php) function & [Kernel::VERSION](https://github.com/symfony/http-kernel/blob/master/Kernel.php#L66) constant.

What do you think ?

Command/DumpSitemapsCommand.php
```diff
@@ -55,7 +55,7 @@ protected function configure()
                 'target',
                 InputArgument::OPTIONAL,
                 'Location where to dump sitemaps. Generated urls will not be related to this folder.',
-                'web'
+                'public'
```
gulaandrij: \App\Kernel::MAJOR_VERSION >= 4 ? 'public' : 'web'

yann-eugone: I cannot agree with this, because it assume that `App\Kernel` class exists.
Outside a Symfony project (during testing for instance) it will fail.

yann-eugone:
Moved to a PR of mine, please see #172

PR #175, opened 2019-02-14
Add autowiring alias for DumperInterface

norkunas:
I think it's a common case when you want to trigger sitemap dump in another context and this will leverage the autowiring

### DIFF
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -35,6 +35,8 @@
         <service id="presta_sitemap.dump_command" class="Presta\SitemapBundle\Command\DumpSitemapsCommand" public="true">
             <tag name="console.command"/>
         </service>
+
+        <service id="Presta\SitemapBundle\Service\DumperInterface" alias="presta_sitemap.dumper_default" />
     </services>
 
 </container>
--- a/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
+++ b/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Presta\SitemapBundle\Test\DependencyInjection;
+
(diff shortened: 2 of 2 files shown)


yann-eugone:
Hi @norkunas, thank you for this suggestion.

But I'm afraid that this won't work in all supported versions of this bundle : FQCN as service id was made possible in the same time as services auto-registering : as of 3.3. 

Can you please : 
* add this alias only for symfony >= 3.3
* add some tests

norkunas:
Are you sure that FQCN doesn't work in olders versions? I didn't find any materials where it states that FQCN are forbbiden. 3.3 only implemented feature which allows to skip `class` definition

yann-eugone:
I pretty sure it is, but I had some difficulties to find a proof.

The closest thing I've found is https://github.com/symfony/symfony/issues/21193

I did the test in a brand new 2.8 application. The container build successfully with the alias you defined.

BUT, the service is lowercased... In our case, it doesn't matter. Will merge

norkunas:
@yann-eugone i've added a test, but not sure if this is the right way to do this.

norkunas:
Also locally i'm getting this:
```
PHP Fatal error:  Cannot redeclare static Symfony\Bundle\FrameworkBundle\Test\WebTestCase::$container as non static Presta\SitemapBundle\Tests\Command\DumpSitemapsCommandTest::$container in /home/norkunas/www/PrestaSitemapBundle/Tests/Command/DumpSitemapsCommandTest.php on line 29
```

probably tests should be updated to correctly work in newer php versions?

norkunas:
ping

yann-eugone:
Where is this error encountered ? I can't see anything in the Travis builds.

norkunas:
I'm on php 7.3, maybe thats the reason

yann-eugone:
Ok, this is why. I will have a look on this failure within the day.

PR #181, opened 2019-03-22
Dispatch event with correct section when requesting section

ostrolucky:
We use your SitemapController and also listen to your event so we can process your Event. Problem is that section of this event is never filled. We want to avoid regenerating whole sitemap again when requesting single section only.

PHP requirement increased because I need $this in closure which is not supported in PHP 5.3 and you don't test for <5.6 anyway.

### DIFF
--- a/Service/Generator.php
+++ b/Service/Generator.php
@@ -90,7 +90,7 @@ class Generator extends AbstractGenerator implements GeneratorInterface
             return $this->cache->fetch($name);
         }
 
-        $this->generate();
+        $this->populate($name);
 
         if ('root' == $name) {
             return $this->getRoot();
--- a/Tests/Service/GeneratorTest.php
+++ b/Tests/Service/GeneratorTest.php
@@ -11,9 +11,11 @@
 
 namespace Presta\SitemapBundle\Test\Sitemap;
 
+use Presta\SitemapBundle\Event\SitemapPopulateEvent;
 use Presta\SitemapBundle\Service\Generator;
 use Presta\SitemapBundle\Sitemap;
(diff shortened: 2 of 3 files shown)


yann-eugone:
We removed php < 5.6 from pipeline, but introducing a major version just for a test trick seems overkill to me.

Maybe you can just perform the assertion after the event dispatch, the same way you did against the var you are using to test whether or not your listener was called

ostrolucky:
This package definitely doesn't support PHP 5.3, it's using eg short arrays here https://github.com/ostrolucky/PrestaSitemapBundle/blob/565725813d5d3eec206e99f0876483218d9a3dbe/Sitemap/Url/UrlConcrete.php#L118 so I've just fixed two issues at the same time.

ostrolucky:
ping @yann-eugone Can we merge this soon? Or do you really want me to make this PHP 5.3 compatible, even though this bundle doesn't work with PHP 5.3?

yann-eugone:
We can merge it soon, but I will not tag it soon.
Updating PHP minimum version mean doing a major version. I really want to do it, but if I do, I will require higher versions of everything, and merge / develop more features that was stashed for this next major.

ostrolucky:
If that's your biggest pet peeve, I can just revert composer.json change. We are not bumping requirement, we are fixing it. This bundle does not work in php 5.3

yann-eugone:
Ok, my bad on this. You're right, `UrlConcrete` is maybe the most used class.
I think that phpstorm replaced it, and any reviewer has seen it on time.
I will tag an `1.6.0`. Anyway, I hope people don't use PHP 5.3 anymore...

PR #184, opened 2019-04-01
 Solve Symfony 4.x deprecations 

ostrolucky:
Didn't require too much more changes after yours. Basic was to get rid of subclassing main Controller. This is kinda BC break in case somebody subclasses controller, but no way around that.


yann-eugone:
Hi @ostrolucky . Thank you for this, it will help.
Those that rely on this controller are doing it wrong, but yeah, this is BC.
The only thing that makes me a bit scared is the routing controller syntax, I can't remember if this is something that was introduced in a later version of the Framework.

Anyway, I think it is time for this bundle to drop some versions support, will try to work on this next week, so yes, we will merge this for the next major.

ostrolucky:
Well tests should cover that and 2.8 is passing

norkunas:
Is there anything more to do? :)

yann-eugone:
This pull request contains BC. We should merge it before next major version.

norkunas:
Yes, the question was meant to be is there anything to do before releasing next major? :)

yann-eugone:
Not sure. I had no plan for that.
Maybe we can end #202 and tag the last minor release.

PR #194, opened 2019-04-15
Fix default target directory implementation

kiler129:
The https://github.com/prestaconcept/PrestaSitemapBundle/pull/192 added a new option to configure default target directory. However, the implementation was slightly wrong: `Command::configure()` is called during `Command::construct()` which meant the default target directory was always set to `null` ;)

cc @yann-eugone


yann-eugone:
Nice catch

PR #202, opened 2019-06-12
#201: Add DateTimeInterface support

### LINKED ISSUE #201: [FR] Support For DateTimeImmutable
### Problem

Running the bundle on 7.3 I discovered that last update date cannot be set to an instance of `\DateTimeImmutable`. The problem is caused by the typehint on `UrlConcrete`:

https://github.com/prestaconcept/PrestaSitemapBundle/blob/5ec8028a16c63c97ab328f7acf456eac571916bc/Sitemap/Url/UrlConcrete.php#L89-L94

### Solution
The proper typehint would be `\DateTimeInterface`, but since the bundle supports PHP 5 I suggest that the setter is changed to manually trigger the same messages as native typehint:

```php
    /**
     * @param DateTimeInterface|DateTime|null $lastmod
     *
     * @return UrlConcrete
     */
    public function setLastmod($lastmod = null)
    {
        if (!($lastmod instanceof \DateTimeInterface || $lastmod instanceof \DateTime)) {
            $typePassed = is_object($lastmod) ? \get_class($lastmod) : \gettype($lastmod);
            
            if (\PHP_MAJOR_VERSION >= 7) {
                throw new \TypeError('Argument 1 passed to ' . __METHOD__ . "() must be an instance of DateTimeInterface, '$typePassed' given");
            } else {
                \trigger_error('Argument 1 passed to ' . __METHOD__ . "() must be an instance of DateTime, '$typePassed' given", E_USER_ERROR);
                
                return $this;
            }
        }
        
        $this->lastmod = $lastmod;

        return $this;
    }
```

### Backward compatibility
To compare the behavior:
 - **Old:** https://3v4l.org/kgBOL
 - **New:** https://3v4l.org/NHDnA

That way there is no difference in behavior of invalid arguments, but on PHP 7.3 it's possible to pass `\DateTimeImmutable`.

WDYT about such change? It will greatly simplify our workflow (so we don't have to convert the dates for every link).

kiler129:
As discussed in #201 I implemented support for `DateTimeInterface` for the `UrlConcrete`. I also found that `DateTime` was used in Google News & Google Video sitemaps - I fixed only the first one, since GVideo is long time gone ;)

Closes #201

### DIFF
--- a/Sitemap/Url/UrlConcrete.php
+++ b/Sitemap/Url/UrlConcrete.php
@@ -12,6 +12,7 @@
 namespace Presta\SitemapBundle\Sitemap\Url;
 
 use DateTime;
+use DateTimeInterface;
 use Presta\SitemapBundle\Sitemap\Utils;
 
 /**
@@ -36,7 +37,7 @@ class UrlConcrete implements Url
     protected $loc;
 
     /**
-     * @var DateTime|null
+     * @var DateTimeInterface|DateTime|null
      */
     protected $lastmod;
 
@@ -53,12 +54,12 @@ class UrlConcrete implements Url
     /**
      * Construct a new basic url
      *
-     * @param string        $loc        Absolute url
-     * @param DateTime|null $lastmod    Last modification date
-     * @param string|null   $changefreq Change frequency
-     * @param float|null    $priority   Priority
+     * @param string                          $loc        Absolute url
+     * @param DateTimeInterface|DateTime|null $lastmod    Last modification date
+     * @param string|null                     $changefreq Change frequency
+     * @param float|null                      $priority   Priority
      */
-    public function __construct($loc, DateTime $lastmod = null, $changefreq = null, $priority = null)
+    public function __construct($loc, $lastmod = null, $changefreq = null, $priority = null)
     {
         $this->setLoc($loc);
         $this->setLastmod($lastmod);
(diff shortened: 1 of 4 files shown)


yann-eugone:
Saying DateTimeInterface was introduced in `php 5.5` do not mean that I want you to bump the minimum php version of this bundle (next major will require `php >=7.1`).
I'm sorry about this misunderstanding.

The first solution you proposed in the issue seems good to me.,
You can just check the existence of the interface : `interface_exists('DateTimeInterface')` instead of checking PHP major version.

kiler129:
@yann-eugone Anything I can do more to make this mergable? :)

Sitemap/Url/UrlConcrete.php
```diff
@@ -87,19 +88,38 @@ public function getLoc()
...
+            $type = is_object($lastmod) ? \get_class($lastmod) : \gettype($lastmod);
+
+            if (\PHP_MAJOR_VERSION >= 7) {
+                throw new \TypeError(
+                    'Argument 1 passed to ' . __METHOD__ .
+                    "() must be an instance of DateTimeInterface or null, '$type' given"
+                );
+            }
+
+            \trigger_error(
```
yann-eugone:
```
            $expectedType = \version_compare(\PHP_VERSION, '5.5.0', '>=') ? 'DateTimeInterface' : 'DateTime'
            \trigger_error(
                'Argument 1 passed to ' . __METHOD__ . "() must be an instance of $expectedType or null, '$type' given",
                \E_USER_ERROR
            );
```

yann-eugone:
I was wondering my thought were understandable, so i started a review, is it more accurate now ?

yann-eugone:
Did it in #210 
Thank you anyway

PR #204, opened 2019-06-27
Fix event deprecation

### LINKED ISSUE #203: SitemapPopulateEvent extends deprecated Event class
### Problem
Since Symfony 4.3 the `Symfony\Component\EventDispatcher\Event` is deprecated. It should extend `Symfony\Contracts\EventDispatcher\Event`.

Due to the ancient version constrain in this bundle I think this issue needs to target next major.

### Reasoning
This causes a direct deprecation notice in tests when even is used like so:
```
Remaining direct deprecation notices (1)

  1x: The "Presta\SitemapBundle\Event\SitemapPopulateEvent" class extends "Symfony\Component\EventDispatcher\Event" that is deprecated since Symfony 4.3, use "Symfony\Contracts\EventDispatcher\Event" instead.
    1x in MyInternalClassTest::testMyTest from App\Tests\EventListener\Foo\Bar
```

Additionally next version of Symofny will remove the old class.

### Workaround
The deprecation can be shifted from direct to indirect by creating mocks of the `SitemapPopulateEvent` instead, like so:

```php
<?php
declare(strict_types=1);

namespace App\Tests\Fixtures\External\Presta;

use Presta\SitemapBundle\Event\SitemapPopulateEvent;
use Presta\SitemapBundle\Service\UrlContainerInterface;
use Prophecy\Prophet;

final class FakeSitemapPopulateEvent
{
    /**
     * @var Prophet|null
     */
    private static $prophet;

    public static function createMock(
        UrlContainerInterface $urlContainer,
        $section = null
    ): SitemapPopulateEvent {
        if (self::$prophet === null) {
            self::$prophet = new Prophet();
        }

        $prophecy = self::$prophet->prophesize(SitemapPopulateEvent::class);
        $prophecy->getUrlContainer()->willReturn($urlContainer);
        $prophecy->getSection()->willReturn($section);

        return $prophecy->reveal();
    }

    private function __construct()
    {
    }
}
```

While usages in tests needs to be modified:
```php
//$ev = new SitemapPopulateEvent($container->reveal(), 'unknown-section');
  $ev = FakeSitemapPopulateEvent::createMock($container->reveal(), 'unknown-section');
```

norkunas:
Closes #203

### DIFF
--- a/Event/SitemapPopulateEvent.php
+++ b/Event/SitemapPopulateEvent.php
@@ -12,68 +12,135 @@
 namespace Presta\SitemapBundle\Event;
 
 use Presta\SitemapBundle\Service\UrlContainerInterface;
-use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\EventDispatcher\Event as BaseEvent;
+use Symfony\Contracts\EventDispatcher\Event as ContractsBaseEvent;
 
-/**
- * Manage populate event
- *
- * @author depely
- */
-class SitemapPopulateEvent extends Event
-{
+if (is_subclass_of('Symfony\Component\EventDispatcher\EventDispatcher', 'Symfony\Contracts\EventDispatcher\EventDispatcherInterface')) {
     /**
-     * @Event("Presta\SitemapBundle\Event\SitemapPopulateEvent")
-     */
-    const ON_SITEMAP_POPULATE = 'presta_sitemap.populate';
-
-    /**
-     * @var UrlContainerInterface
-     */
-    protected $urlContainer;
-
-    /**
-     * Allows creating EventListeners for particular sitemap sections, used when dumping
-     * @var string
-     */
-    protected $section;
-
-    /**
-     * @param UrlContainerInterface $urlContainer
-     * @param string|null           $section
-     */
-    public function __construct(UrlContainerInterface $urlContainer, $section = null)
-    {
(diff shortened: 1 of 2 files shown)


yann-eugone:
I'm not a fan of what Symfony did with these "contracts".
And I'm not a fan of these ugly double classes files that we need to maintain just to avoid these deprecation warning...
For now, I cannot merge this. We need to think about an other solution.

norkunas:
What's wrong with freeing up us from deprecations? :)

2019-07-02, an 21:58, Yann Eugoné <notifications@github.com> rašė:

yann-eugone:
Having deprecation is not a good thing. Agreed.

Fixing it the right way is important too. And I don't think that this kind of trick (several classes in the same file, depending on a condition) help people to understand the code they are using.

So for now, I disagree with this solution.

norkunas:
That's how Symfony solves this in it's codebase, I didn't want to reinvent
the wheel :)

2019-07-02, an 22:21, Yann Eugoné <notifications@github.com> rašė:

yann-eugone:
I know they do... But I'm not comfortable with that, let me some time to find an other solution

norkunas:
Okay, no prob :)

yann-eugone:
Ok, I had some words with colleagues, everyone agree that this is ugly, but no one is able to provide a better solution. So... Let's merge it :)

PR #205, opened 2019-07-11
Auto adding alternales

fabianoroberto:
Handle auto adding alternales by adding `defaults.default_locale` and `default.locales` parameters

Also added `.idea` in `.gitignore` file to prevent commit unuseful files of Jetbrains IDE.

Any dubts/suggestions about PR are welcome


yann-eugone:
Thank you for this idea.
We will need more explanation about what you coded, because there is a lot, and i'm not sure to understand every part.

Can you please add some explanation comments, and write some tests to cover your changes ?

.gitignore
```diff
@@ -4,3 +4,4 @@ composer.lock
 phpunit.xml
 vendor/
 Tests/web
+.idea/
```
yann-eugone: This is very personal, and should be in you global settings

DependencyInjection/Configuration.php
```diff
@@ -69,6 +69,18 @@ public function getConfigTreeBuilder()
                         ->scalarNode('priority')->defaultValue(1)->end()
                         ->scalarNode('changefreq')->defaultValue(UrlConcrete::CHANGEFREQ_DAILY)->end()
                         ->scalarNode('lastmod')->defaultValue('now')->end()
+                        ->scalarNode('default_locale')
```
yann-eugone: I do not think that this should belongs to `defaults`, maybe a `canBeEnabled` distinct section

EventListener/RouteAnnotationEventListener.php
```diff
@@ -47,14 +48,30 @@ class RouteAnnotationEventListener implements EventSubscriberInterface
...
      * @param RouterInterface $router
      * @param string          $defaultSection
+     * @param array           $defaultOptions
      */
-    public function __construct(RouterInterface $router, $defaultSection)
-    {
+    public function __construct(
+        RouterInterface $router,
+        $defaultSection,
+        $defaultOptions = [
```
yann-eugone: Please don't.
Code should not change that much at once.

EventListener/RouteAnnotationEventListener.php
```diff
@@ -96,6 +113,14 @@ private function addUrlsFromRoutes(SitemapPopulateEvent $event)
                 continue;
             }
 
+            if ($this->defaultOptions['default_locale']) {
+                if (strpos($name, $this->defaultOptions['default_locale']) === false) {
+                    continue;
+                }
+
+                $name = preg_replace('/[a-z]+__RG__/', '', $name);
```
yann-eugone: I'm not sure to understand this part, what's this for ?

EventListener/RouteAnnotationEventListener.php
```diff
@@ -195,12 +217,30 @@ public function getOptions($name, Route $route)
     protected function getUrlConcrete($name, $options)
     {
         try {
-            return new UrlConcrete(
-                $this->getRouteUri($name),
+            $params = [];
```
yann-eugone: The method is called `getUrlConcrete`, maybe we can stick to that for now, but decorate that url later ?

EventListener/RouteAnnotationEventListener.php
fabianoroberto: I'm just adding optional $defaultOptions argument, but row was too long

EventListener/RouteAnnotationEventListener.php
fabianoroberto: I'm use http://jmsyst.com/bundles/JMSTranslationBundle to map different url per language. New url have that structure

 it__RG__landing_home       /home                                                       
  en__RG__landing_home    /en/home                                                    
  it__RG__landing_about      /chi-siamo                                                  
  en__RG__landing_about    /en/about-us                                                
  it__RG__landing_agency    /cosa-facciamo

So to obtain "canonical_url" trim first part of route name

EventListener/RouteAnnotationEventListener.php
fabianoroberto: You right, maybe is better introduce a new method to handle this case

DependencyInjection/Configuration.php
fabianoroberto: I try to rewrite this part

EventListener/RouteAnnotationEventListener.php
yann-eugone: So, the Pull Request you wrote is about supporting translatable URLs **with** `JMSTranslationBundle`.

But what about Symfony's I18n routes ?
- https://symfony.com/doc/current/routing.html#localized-routes-i18n
- https://symfony.com/blog/new-in-symfony-4-1-internationalized-routing ?

fabianoroberto: Yep, i'll try to support also symfony i18n, but it isn't supported on symfony 3.4 https://symfony.com/doc/3.4/routing.html#translating-routes

fabianoroberto: I handled both cases in my last commit

DependencyInjection/PrestaSitemapExtension.php
```diff
@@ -39,6 +39,10 @@ public function load(array $configs, ContainerBuilder $container)
         $container->setParameter($this->getAlias() . '.defaults', $config['defaults']);
         $container->setParameter($this->getAlias() . '.default_section', $config['default_section']);
 
+        if ($this->isConfigEnabled($container, $config['alternate'])) {
```
yann-eugone: Otherwhise, the parameter is not set, which will cause errors : missing parameter : https://travis-ci.org/prestaconcept/PrestaSitemapBundle/builds/592629655

DependencyInjection/Configuration.php
```diff
@@ -78,6 +79,37 @@ public function getConfigTreeBuilder()
...
+                                ->then(function ($v) { return preg_split('/\s*,\s*/', $v); })
+                            ->end()
+                            ->prototype('scalar')->end()
+                            ->info('Array of locales to generate alternate (hreflang) urls')
+                        ->end()
+                        ->enumNode('i18n')
+                            ->defaultValue('symfony')
+                            ->values(['symfony', 'jms'])
+                            ->info('Name of project bundle to create i18n routes. Possible values are symfony or jms')
+                    ->end()
```
yann-eugone: We may be missing an `end()` (or the indentation is wrong)

EventListener/RouteAnnotationEventListener.php
```diff
@@ -47,14 +48,20 @@ class RouteAnnotationEventListener implements EventSubscriberInterface
...
+     * @var array
+     */
+    private $alternateSection;
+
     /**
      * @param RouterInterface $router
      * @param string          $defaultSection
      */
-    public function __construct(RouterInterface $router, $defaultSection)
+    public function __construct(RouterInterface $router, $defaultSection, $alternateSection)
```
yann-eugone: Please type the parameter

EventListener/RouteAnnotationEventListener.php
```diff
@@ -214,6 +274,53 @@ protected function getUrlConcrete($name, $options)
...
+
+                foreach ($options['locales'] as $locale) {
+                    $params['_locale'] = $locale;
+
+                    $url->addLink($this->getRouteUri($name, $params), $locale);
+                }
+            }
+
+            return $url;
+        } catch (\Exception $e) {
```
yann-eugone: Variables should never be that short, please consider using `$exception` instead

EventListener/RouteAnnotationEventListener.php
```diff
@@ -214,6 +274,53 @@ protected function getUrlConcrete($name, $options)
...
+            throw new \InvalidArgumentException(
+                sprintf(
+                    'Invalid argument for route "%s": %s',
```
yann-eugone: All exception will cause that message, are you sure that this is not too specific ?

EventListener/RouteAnnotationEventListener.php
```diff
@@ -75,7 +82,11 @@ public function registerRouteAnnotation(SitemapPopulateEvent $event)
         $section = $event->getSection();
 
         if (is_null($section) || $section === $this->defaultSection) {
-            $this->addUrlsFromRoutes($event);
+            if ($this->alternateSection) {
```
yann-eugone: This is too much duplicated code for me, we should just be able to decorate the URL in `addUrlsFromRoutes` this will avoid duplicating the whole logic in two separated methods

fabianoroberto: Removed method to prevent duplicate code

EventListener/RouteAnnotationEventListener.php
fabianoroberto: Done

DependencyInjection/Configuration.php
fabianoroberto: Thanks, my mistake

DependencyInjection/PrestaSitemapExtension.php
fabianoroberto: fixed

Resources/config/route_annotation_listener.xml
```diff
@@ -11,6 +11,7 @@
         <service id="presta_sitemap.eventlistener.route_annotation" class="%presta_sitemap.eventlistener.route_annotation.class%">
             <argument type="service" id="router"/>
             <argument>%presta_sitemap.default_section%</argument>
+            <argument type="collection">%presta_sitemap.alternate%</argument>
```
l-vo: Why `type=collection` ? Shouldn't it be simply `<argument>%presta_sitemap.alternate%</argument>` ?

DependencyInjection/PrestaSitemapExtension.php
l-vo: Not sure it is fixed, when the config is disabled (`enabled: false`), the parameter doesn't exist too and the container compilation fails. Maybe set `presta_sitemap.alternate` as an empty array when the section is disabled ?

Resources/config/route_annotation_listener.xml
fabianoroberto: Because is an array of elements like this:

```yaml
 presta_sitemap:
     alternate:
         default_locale: 'en'
         locales: ['en', 'it']
         i18n: jms
 ```

I'm not sure is the same use `<argument>`

Comment (yann-eugone) [reply Resources/config/route_annotation_listener.xml]: @l-vo suggestion would work, but I prefer to see in the service declaration that this argument is an array, so it's ok for me

File: DependencyInjection/PrestaSitemapExtension.php (hunk 6 above)
Comment (yann-eugone) [reply DependencyInjection/PrestaSitemapExtension.php]: We should add a use case for this.
Running this bundle with that feature disabled should be tested IMO

File: Resources/config/route_annotation_listener.xml (hunk 12 above)
Comment (l-vo) [reply Resources/config/route_annotation_listener.xml]: Hum, does it work like that on your environments ? To me use `type="collection"` results to parse child nodes of ` <argument type="collection">` (see  [XmlFileLoader](https://github.com/symfony/symfony/blob/33e2735d1409554a040b3a293ee8e121c84248cb/src/Symfony/Component/DependencyInjection/Loader/XmlFileLoader.php#L500)). But in this case there is no child node. So it would lead to receive an empty array as alternate argument; but maybe I miss something ?

Comment (yann-eugone) [reply Resources/config/route_annotation_listener.xml]: Using `<argument type="collection">` is only relevant when you are planning to define the collection structure manually : 

```
<argument type="collection">
    <argument key="foo">FOO</argument>
    <argument key="bar">BAR</argument>
</argument>
```

In our case, the value is injected from a parameter, there is no need for the `type="collection"`, I still encourage to do it, for the sake of maintenability.

Comment (l-vo) [reply Resources/config/route_annotation_listener.xml]: In this case, to me, it's not a matter of tase, this just doesn't work. Sorry for the noise if I'm wrong 🙂

Comment (yann-eugone) [reply Resources/config/route_annotation_listener.xml]: You tested it and it fails, or your did not tested it and you are expecting it to fail ?
Can't find any example in my memory, but I feel like I did it many times without any problem...

Comment (l-vo) [reply Resources/config/route_annotation_listener.xml]: I tested id but maybe I made a configuration mistake; it's why I didn't affirm I tested it. It's not a problem if the argument is empty (like `<argument type="collection" />`) and you replace it in a compiler pass, but in that case the parameter seems not present when the service configuration xml is loaded.

Comment (yann-eugone) [reply Resources/config/route_annotation_listener.xml]: If the parameter is not defined, no matter the structure of XML service configuration, the compilation will fail, because of that missing parameter

File: DependencyInjection/PrestaSitemapExtension.php (hunk 6 above)
Comment (fabianoroberto) [reply DependencyInjection/PrestaSitemapExtension.php]: I'll try to create a use case and a test soon

**Comment by fabianoroberto** [comment]
I added a simple test to check alternate section. If you merge before #224 I can add other integration tests

File: Tests/DependencyInjection/PrestaSitemapExtensionTest.php [hunk 13]
```diff
@@ -17,4 +17,31 @@ public function testDumperAliasIsSet()
 
         self::assertTrue($containerBuilder->hasAlias('Presta\SitemapBundle\Service\DumperInterface'));
     }
+
+    public function testAlternate()
```
yann-eugone: This is indeed the good direction, but we must add more tests for that section : maybe a data provider with several use cases : 
- no config for the bundle
- disabled alternate section
- enabled alternate section with config

yann-eugone:
Hi @fabianoroberto !
Since the test suite refactoring has been merged, do you still want to work on this ? Or do you want someone to continue your work ?

fabianoroberto:
Hi @yann-eugone, at this moment I'm a little busy so I cannot follow this task. I'd be very happy if someone can continue this MR

yann-eugone:
Replaced with #251

PR #215, opened 2019-10-22
Fix nonexistent sitemap

pulzarraider:
This PR fixes the problem that occurs in our project.

When new sitemaps are created, they are deleted from target dir and than copying from the tmp dir. In this time, when the Googlebot comes for the sitemap, the sitemap is missing for a while and it gets error 404.

This PR fixes the problem with small tweak - only nonexistent sitemaps will be deleted before mirroring the tmp dir.


yann-eugone:
Hello. Thank you for using this bundle.

I'm sorry, but I understand neither your problem nor the solution you proposed.

Can you please try to rewrite the description in a more detailed way ?

pulzarraider:
@yann-eugone we are generating news sitemaps with this bundle. Googlebot is downloading the sitemaps from us very often. But some requests fail with error 404. We were investigating the problem and we have found it during generating the sitemaps. The new sitemap is generating in /tmp dir what is great, but before the /tmp content is copying to target, the original (old) sitemap is generating. It means there is small amount of time, when the sitemap does not exist.

pulzarraider:
@yann-eugone is my description more clear?

yann-eugone:
Let me try to rewrite what I understand about your problem : 

You sometime have some `404 Not Found` with some part of your sitemap.
You've find out that this is because the bot is asking for it during the sitemap generation.

Let's having a look to the [Dumper](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php) service : 

- The sitemap is generated in a [tmp directory](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L128)
- When generation is completed, the new sitemap is [activated](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L196)

The current sitemap is not removed during generation, it is removed during the "activation" step : 

- [delete existing sitemap](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L208)
- [mirror temporary dir to sitemap dir](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L211)
- [remove temporary dir](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L212)

So your problem occurs between the 2 first instruction of the activation step : delete -> mirror

Am I right ?

-----

About the code you added. From now :

- we will only remove the files that are no longer part of the sitemap
- the [filesystem::mirror](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Service/Dumper.php#L211) with override option will just replace old sitemap with new ones

Is it also right ?

pulzarraider:
> So your problem occurs between the 2 first instruction of the activation step : delete -> mirror
> Am I right ?

Yes, you're right.

After applying fix from this PR the occasional `404 Not found` errors have disappeared.

yann-eugone:
We will need more tests about this. Can you add a test case to handle it ?

yann-eugone:
Asking for tests about this one was a hard task. I've wondered what kind of test I could do, without being able to find anything : the thing that an assertion must be done during the test itself is a PITA...

I decided to merge it (almost) as is. Sorry for the delay

PR #220, opened 2020-01-05
Fixes PS4 tests autoloading

NoodlesNZ:
Composer 1.10+ complains that the test classes do not comply with psr-4 autoloading standard and won't be loaded in composer 1.11.

This can be replicated by updating composer to the latest snapshot (`composer self-update --snapshot`) and then running `composer install -o`

Looking at official Symfony modules, they use this method to fix the issue.

### DIFF
--- a/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
+++ b/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Presta\SitemapBundle\Test\DependencyInjection;
+namespace Presta\SitemapBundle\Tests\DependencyInjection;
 
 use PHPUnit\Framework\TestCase;
 use Presta\SitemapBundle\DependencyInjection\PrestaSitemapExtension;
--- a/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
+++ b/Tests/DependencyInjection/PrestaSitemapExtensionTest.php
@@ -6,7 +6,7 @@ use PHPUnit\Framework\TestCase;
 use Presta\SitemapBundle\DependencyInjection\PrestaSitemapExtension;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
-class RouteAnnotationEventListenerTest extends TestCase
+class PrestaSitemapExtensionTest extends TestCase
 {
     public function testDumperAliasIsSet()
     {
--- a/Tests/EventListener/RouteAnnotationEventListenerTest.php
+++ b/Tests/EventListener/RouteAnnotationEventListenerTest.php
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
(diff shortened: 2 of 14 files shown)


yann-eugone:
This is indeed a problem. I don't think there is any risk with this add, as tests should event not be included in a production environment.

Still, I was wondering about those deprecations, so I did reproduce the steps you noted.
It appear that we really has a problem with our tests :

Some classes are using `Presta\SitemapBundle\Test\` instead of `Presta\SitemapBundle\Tests\` and **this is why we have deprecations** !

Can you fix these classes namespace in that PR too ?

NoodlesNZ:
Sure thing, I'll do that now

NoodlesNZ:
Fixing the namespace `Test` -> `Tests` gets rid of most of the problems, but there's a few left over where the class/namespace don't match the filename/path. e.g.

```
Deprecation Notice: Class Presta\SitemapBundle\Tests\Sitemap\GeneratorTest located in ./Tests/Service/GeneratorTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v1.11+. in phar:///usr/local/Cellar/composer/1.9.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:185
Deprecation Notice: Class Presta\SitemapBundle\Tests\DependencyInjection\RouteAnnotationEventListenerTest located in ./Tests/DependencyInjection/PrestaSitemapExtensionTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v1.11+. in phar:///usr/local/Cellar/composer/1.9.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:185
Deprecation Notice: Class Presta\SitemapBundle\Tests\Sitemap\RouteAnnotationEventListenerTest located in ./Tests/EventListener/RouteAnnotationEventListenerTest.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v1.11+. in phar:///usr/local/Cellar/composer/1.9.1/bin/composer/src/Composer/Autoload/ClassMapGenerator.php:185
```

NoodlesNZ:
I can either fix up the filename/paths to match the namespace or fix the namespace to match the filename/path? Which would you prefer?

yann-eugone:
IMO we must follow the standard style : same namespace as the class it tests : 

```
Presta\SitemapBundle\Service\Generator => Presta\SitemapBundle\Tests\Service\GeneratorTest
Presta\SitemapBundle\DependencyInjection\PrestaSitemapExtension => Presta\SitemapBundle\Tests\DependencyInjection\PrestaSitemapExtensionTest
Presta\SitemapBundle\EventListener\RouteAnnotationEventListener => Presta\SitemapBundle\Tests\EventListener\RouteAnnotationEventListenerTest
```

Then, the path should match PSR4 rules

NoodlesNZ:
Did you want me to squash the commits?

yann-eugone:
No, GitHub will do this for us ! 

Thank you for your help !

Issue #221, opened 2020-01-10
The controller for URI "/sitemap.xml" is not callable. Class "presta_sitemap.controller" does not exist.


ostrolucky:
This is happens with update from 1.7 to 2.0. Problem seems to be with using `presta_sitemap.controller::indexAction` instead of `presta_sitemap.controller:indexAction`. Running Symfony 3.4 here.

https://symfony.com/blog/new-in-symfony-4-1-deprecated-the-bundle-notation

> we noticed a related inconsistency that could be fixed. When defining controllers as services, you must use a single colon (:) instead of a double colon (::) to separate the service ID and the method name. In other parts of Symfony you always use a double colon (::) to separate classes and method names, so this difference is confusing and it complicates the learning curve for no real benefit. That's why in Symfony 4.1 you can always use a double colon to separate the method names, even for controllers as services.

So using double colon there breaks all < symfony 4.1 applications

yann-eugone:
As promised, I did run some tests on brand new Symfony projects, using last stable version of this bundle.
But I noticed no error on any version :
- 3.4.36
- 4.1.12
- 4.2.12
- 4.3.9
- 4.4.2
- 5.0.2

Did I miss something ?

ostrolucky:
This happens when accessing `/sitemap.xml` url

yann-eugone:
I know, but as I said I did no reproduced your problem with my symfony 3.4 test project.
Is there any chance that symfony fixed it the version I'm using ?

ostrolucky:
```
➜  PhpstormProjects symfony new --version=3.4 presta-test
* Creating a new Symfony 3.4 project with Composer
  (running /usr/local/bin/composer create-project symfony/skeleton /Users/gostrolucky/PhpstormProjects/presta-test 3.4.*)

* Setting up the project under Git version control
  (running git init /Users/gostrolucky/PhpstormProjects/presta-test)


 [OK] Your project is now ready in /Users/gostrolucky/PhpstormProjects/presta-test


➜  PhpstormProjects cd presta-test
➜  presta-test git:(master) hstr
composer require presta/sitemap-bundle
➜  presta-test git:(master) composer require presta/sitemap-bundle
Using version ^2.0 for presta/sitemap-bundle
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Restricting packages listed in "symfony/symfony" to "3.4.*"
Package operations: 1 install, 0 updates, 0 removals
[...]

PR #222, opened 2020-01-10
Fix invalid routing config on Symfony < 4.1

ostrolucky:
Fixes https://github.com/prestaconcept/PrestaSitemapBundle/issues/221. Was regression from https://github.com/prestaconcept/PrestaSitemapBundle/pull/213

### DIFF
--- a/Resources/config/routing.yml
+++ b/Resources/config/routing.yml
@@ -1,11 +1,11 @@
 PrestaSitemapBundle_index:
     path:     "/%presta_sitemap.sitemap_file_prefix%.{_format}"
-    defaults: { _controller: presta_sitemap.controller::indexAction }
+    defaults: { _controller: Presta\SitemapBundle\Controller\SitemapController::indexAction }
     requirements:
         _format: xml
 
 PrestaSitemapBundle_section:
     path:     "/%presta_sitemap.sitemap_file_prefix%.{name}.{_format}"
-    defaults: { _controller: presta_sitemap.controller::sectionAction }
+    defaults: { _controller: Presta\SitemapBundle\Controller\SitemapController::sectionAction }
     requirements:
         _format: xml
--- a/Resources/config/services.xml
+++ b/Resources/config/services.xml
@@ -40,6 +40,7 @@
         </service>
 
         <service id="Presta\SitemapBundle\Service\DumperInterface" alias="presta_sitemap.dumper_default" />
+        <service id="Presta\SitemapBundle\Controller\SitemapController" alias="presta_sitemap.controller" public="true" />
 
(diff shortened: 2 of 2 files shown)


yann-eugone:
Thank you for addind this fix, I feel pretty inconfortable with this issue because we didn't see it before.

Can you imagine a test that did avoid this ?

ostrolucky:
> Can you imagine a test that did avoid this ?

What do you mean?

yann-eugone:
I mean, that issue was not spotted by our test suite.

The usual process is to write a test that fails with the old code source, fix the code so the test pass.

ostrolucky:
I mean no test case was touched in https://github.com/prestaconcept/PrestaSitemapBundle/pull/213 where regression was introduced either. It was also very poor on details. Normally I would adjust tests, but this is not my project and it lacks proper integration tests. Quite contrary, `SitemapControllerTest` goes out of its way to avoid using routing. Routing file for testing was also overriddden for unknown reason. If I was maintainer of this project, I would replace these tests with tests using official practice, by using symfony webtest browser and also remove overridden routes, but I am not so I'm not comfortable outright doing such changes.

yann-eugone:
Fine, I will do my best to test this over all symfony versions we support before merging, releasing it.

I known this package has some troubles, specially with tests. Will do my best to rewrite the test suite some day.

ostrolucky:
Let me know if you need some help. Biggest worry of contributors is that changes are rejected, but if you can guide me what you intend to do, I might help.

yann-eugone:
We will have to merge it if not working on 3.4, but I have to be certain that other versions will not break on the same time. As tests are not helpful (for now) I will have to test it myself, but Monday 😉

yann-eugone:
Did you tested that code against any 4.x version ? On my side I have an other issue with on symfony > 4.0 : 

```
Controller "Presta\SitemapBundle\Controller\SitemapController" cannot be fetched from the container because it is private.
Did you forget to tag the service with "controller.service_arguments"?
```

Caused by 

```
Too few arguments to function Presta\SitemapBundle\Controller\SitemapController::__construct(), 
0 passed in /src/vendor/symfony/http-kernel/Controller/ControllerResolver.php on line 133 
and exactly 2 expected
```

This is because the route is now bound to the `Presta\SitemapBundle\Controller\SitemapController` service id, which is private.

Adding `public="true"` to the alias you introduced solved the issue on my side.

Also, I created #223 with some tests that should prove that this issue will not appear again.

ostrolucky:
Nope, I didn't test it on 4.0, so that error is indeed possible

yann-eugone:
I about to merge this PR, then rebase #223 to see if added tests actually pass with these changes

PR #223, opened 2020-01-13
[WIP] Added test assertions to ensure routing is configured the good way

yann-eugone:
As an addition to #221 & #222 this PR add some assertion that should cause tests to fail at the moment.

When #222 will be merged, this PR will be rebased, and tests should be green again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 matrix:
     include:
-        - php: 7.0
+        - php: 7.1
           env: SYMFONY_VERSION=3.4.*
         - php: 7.4
           env: SYMFONY_VERSION=3.4.*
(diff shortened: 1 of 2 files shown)


ostrolucky:
Why not functional tests like docs suggest? https://symfony.com/doc/current/testing.html#your-first-functional-test It looks less error prone, covers wider array of things and are more concise

yann-eugone:
Because it requires more work, I'm currently working on a test refactor (as promised in #222) this is just a hack that is to be replaced in a few weeks.

yann-eugone:
It was harder than expected (private services access + deprecations) but this is it : 

- Before #222 : https://travis-ci.org/prestaconcept/PrestaSitemapBundle/builds/636442546
- After #222 : https://travis-ci.org/prestaconcept/PrestaSitemapBundle/builds/636773896

PR #226, opened 2020-01-16
Make DumpSitemapsCommand lazy

norkunas:


Command/DumpSitemapsCommand.php
```diff
@@ -56,7 +58,7 @@ public function __construct(RouterInterface $router, DumperInterface $dumper, $d
      */
     protected function configure()
     {
-        $this->setName('presta:sitemaps:dump')
+        $this->setName(self::$defaultName)
```
yann-eugone: Correct me if i'm wrong, but calling `setName()` is not required since `$defaultName` is filled.

norkunas: Yes, you are right, updated :)

PR #227, opened 2020-02-12
Allow multiple videos

l-vo:
Allow to add multiple video for an url. The pattern become the same as used for the images (a `GoogleVideo` object added to the decorator by a `addVideo` method).

Note that a BC layer was added in a separate commit.


Exception/GoogleVideoException.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace Presta\SitemapBundle\Exception;
+
+/**
+ * Exception used when limit is reached on adding video
+ *
+ * @author David Epely <depely@prestaconcept.net>
+ */
+class GoogleVideoException extends GoogleVideoUrlException
```
yann-eugone: The new exception should not inherit from the old one, it should be the opposite, like : https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Environment.php

Exception/GoogleVideoTagException.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace Presta\SitemapBundle\Exception;
+
+/**
+ * Exception used when limit is reached on adding tag to video
+ *
+ * @author David Epely <depely@prestaconcept.net>
+ */
+class GoogleVideoTagException extends GoogleVideoUrlTagException
```
yann-eugone: The new exception should not inherit from the old one, it should be the opposite, like : https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Environment.php

Exception/GoogleVideoUrlException.php
```diff
@@ -15,6 +15,8 @@
  * Exception used when limit is reached on adding video
  *
  * @author David Epely <depely@prestaconcept.net>
+ *
+ * @deprecated Use \Presta\SitemapBundle\Exception\GoogleVideoException instead.
  */
 class GoogleVideoUrlException extends Exception
```
yann-eugone: A `@trigger_error('', E_USER_DEPRECATED);` should be added in `__construct`

Exception/GoogleVideoUrlTagException.php
```diff
@@ -15,6 +15,8 @@
  * Exception used when limit is reached on adding tag to video
  *
  * @author David Epely <depely@prestaconcept.net>
+ *
+ * @deprecated Use \Presta\SitemapBundle\Exception\GoogleVideoTagException instead.
  */
 class GoogleVideoUrlTagException extends Exception
```
yann-eugone: A `@trigger_error('', E_USER_DEPRECATED);` should be added in `__construct`

Sitemap/Url/GoogleVideoUrlDecorator.php
```diff
@@ -198,58 +204,88 @@ class GoogleVideoUrlDecorator extends UrlDecorator
...
+                break;
+            default:
+                $text = 'Using %s is deprecated.';
         }
 
-        parent::__construct($urlDecorated);
+        @trigger_error(
+            sprintf($text, $methodName),
+            E_USER_DEPRECATED
+        );
```
yann-eugone: Can you add `__CLASS__` to all messages, so the FQCN is included in error logs ?

- `Using %s::%s is deprecated. Create a GoogleVideo object instead.`
- `Using %s::%s is deprecated. Retrieve it from GoogleVideo object instead.`
- `Using %s::%s is deprecated. Add it to GoogleVideo object instead.`
- `Using %s::%s is deprecated.`

`@trigger_error(sprintf($text, __CLASS__, $methodName), E_USER_DEPRECATED);`

Tests/Sitemap/Url/GoogleVideoUrlDecoratorTest.php
```diff
@@ -26,6 +29,95 @@ class GoogleVideoUrlDecoratorTest extends TestCase
     protected $xml;
 
     protected function setUp()
```
yann-eugone: Can you add a test case for a video url with multiple video object, with some assertions on the generated xml ?

Exception/GoogleVideoException.php
l-vo: I'm not agree with that, this example is IMHO not ok when we are talking about exceptions. The new code throws a `GoogleVideoException`, but the application may try to catch the old exception `GoogleVideoUrlException`. If `GoogleVideoUrlException` extends from `GoogleVideoException` and the new code throws a `GoogleVideoException`, exception won't be caught by the application. Are you ok with that ?

l-vo: I'm not agree with that, this example is IMHO not ok when we are talking about exceptions. The new code throws a `GoogleVideoException`, but the application may try to catch the old exception `GoogleVideoUrlException`. If `GoogleVideoUrlException` extends from `GoogleVideoException` and the new code throws a `GoogleVideoException`, exception won't be caught by the application. Are you ok with that ?

Exception/GoogleVideoUrlException.php
l-vo: If we keep the inheritance above (`GoogleVideoException` extends `GoogleVideoUrlException`), we won't be able to trigger a deprecation in the constructor (since the depreciation would be triggered when the new exception is thrown). I think we can't preserve try/catch compatibility AND trigger a deprecation. Let me know if you prefer to not have a BC layer for exception (or to keep old exceptions).

l-vo:
@yann-eugone I added some tests and change deprecation messages. For the others feedbacks, please consider my answers :)

Exception/GoogleVideoException.php
yann-eugone: OK. You convinced me. Let's try it like that.

Exception/GoogleVideoUrlException.php
yann-eugone: As I said above, I didn't see it like that, but your idea seems good to me.

yann-eugone:
Thank you for this nice addition and your effort with tests & bc layer :)

PR #232, opened 2020-03-01
Allow to set defaults on decorated routes

### LINKED ISSUE #228: Configured defaults (changefreq, priority...) are only added on non-decorated url
Hi :) 

Configured defaults are not added when urls are decorated: https://github.com/prestaconcept/PrestaSitemapBundle/blob/3ca9c8ac03b778674517086d378bdcafbd60a33d/Service/AbstractGenerator.php#L99

Is there any reason for that ? Or is it a bug ?

Thank you for your answer.
Regards.

yann-eugone:
This is supposed to fix #228 

We noticed that `AbstractGenerator` is only setting `changefreq`, `lastmod` & `priority` to instances of `UrlConcrete`.

This is not the expected behavior as all urls can have these xml properties.

With that pull request `AbstractGenerator` will dive in decorated chain of urls until it find an `UrlConcrete` (or not) and set these defaults to that url.


alain-flaus:
:+1:

PR #236, opened 2020-04-03
Typo in "a" instead of "an"

felixprojekt:
"an" should be used before word "interface"


yann-eugone:
Thank you for taking some time with this

PR #237, opened 2020-05-27
Add DumpSitemapMessage for messenger integration

norkunas:
I am not sure which messenger version should be picked?

### DIFF
--- a/DependencyInjection/PrestaSitemapExtension.php
+++ b/DependencyInjection/PrestaSitemapExtension.php
@@ -15,6 +15,7 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
 
 /**
  * This is the class that loads and manages your bundle configuration
@@ -41,6 +42,10 @@ class PrestaSitemapExtension extends Extension
 
         if (true === $config['route_annotation_listener']) {
             $loader->load('route_annotation_listener.xml');
+        }
+
+        if (interface_exists(MessageHandlerInterface::class)) {
+            $loader->load('messenger.xml');
         }
 
         $generator = $container->setAlias('presta_sitemap.generator', $config['generator']);
--- a/Messenger/DumpSitemapMessage.php
+++ b/Messenger/DumpSitemapMessage.php
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * This file is part of the PrestaSitemapBundle package.
+ *
+ * (c) PrestaConcept <www.prestaconcept.net>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Presta\SitemapBundle\Messenger;
+
+/**
+ * Message to dump the sitemaps asynchronously or synchronously in background
+ *
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+class DumpSitemapMessage
+{
+    /**
+     * @var string|null
+     */
+    private $section;
+
+    /**
+     * @var string|null
+     */
+    private $baseUrl;
+
+    /**
+     * @var string|null
+     */
+    private $targetDir;
+
+    /**
+     * @var array
+     */
+    private $options;
+
+    public function __construct(string $section = null, string $baseUrl = null, string $targetDir = null, array $options = [])
+    {
+        $this->section = $section;
+        $this->baseUrl = $baseUrl;
+        $this->targetDir = $targetDir;
+        $this->options = $options;
+    }
+
+    public function getSection(): ?string
+    {
+        return $this->section;
+    }
+
+    public function getBaseUrl(): ?string
+    {
+        return $this->baseUrl;
+    }
+
+    public function getTargetDir(): ?string
+    {
+        return $this->targetDir;
+    }
+
+    public function getOptions(): array
+    {
+        return $this->options;
+    }
+}
--- a/Resources/doc/7-messenger-integration.md
+++ b/Resources/doc/7-messenger-integration.md
@@ -1,0 +1,47 @@
+# Messenger integration
+
+If you have installed [Symfony Messenger](https://symfony.com/doc/current/messenger.html#installation), then you can 
+dispatch `Presta\SitemapBundle\Messenger\DumpSitemapMessage` message to your transport to handle it asynchronously or
+synchronously.
+
+## [Routing the message to your transport](https://symfony.com/doc/current/messenger.html#routing-messages-to-a-transport)
+
+```yaml
+# config/packages/messenger.yaml
+framework:
+    messenger:
+        transports:
+            async: "%env(MESSENGER_TRANSPORT_DSN)%"
+
+        routing:
+            # async is whatever name you gave your transport above
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage':  async
+```
+
+After configuring the message routing dispatch the message like this:
+
+```php
+// src/Controller/DefaultController.php
+namespace App\Controller;
+
+use Presta\SitemapBundle\Messenger\DumpSitemapMessage;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+class DefaultController extends AbstractController
+{
+    public function index(MessageBusInterface $bus)
+    {
+        // this will dispatch to dump all sitemap sections
+        $bus->dispatch(new DumpSitemapMessage());
+
+        // If you wish to dump a single section, change the base url, target dir
+        // and gzip option you can provide these through the message constructor
+        $bus->dispatch(new DumpSitemapMessage('custom_section', 'https://sitemap.acme.org', '/path/to/sitemap', ['gzip' => true]));
+    }
+}
+```
+
+---
+
+« [Dumping sitemap](6-dumping-sitemap.md) • [README](../../README.md) »
--- a/Tests/Integration/config/messenger.yaml
+++ b/Tests/Integration/config/messenger.yaml
@@ -1,0 +1,10 @@
+framework:
+    messenger:
+        transports:
+            async: 'in-memory://'
+        routing:
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage': async
+
+services:
+    Presta\SitemapBundle\Tests\Integration\Controller\MessengerController:
+        tags: ['controller.service_arguments']
--- a/Tests/Integration/src/ContainerConfiguratorTrait.php
+++ b/Tests/Integration/src/ContainerConfiguratorTrait.php
@@ -6,6 +6,7 @@ use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
 use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Messenger\MessageBusInterface;
 
 if (Kernel::VERSION_ID >= 50100) {
     trait ContainerConfiguratorTrait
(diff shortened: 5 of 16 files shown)


yann-eugone:
Hi @norkunas !
First of all, thank you for this nice integration idea.
I'll start having a look at your code today, so I can give you some feedback.

Resources/doc/7-messenger-integration.md
```diff
@@ -0,0 +1,58 @@
...
+// src/Controller/DefaultController.php
+namespace App\Controller;
+
+use Presta\SitemapBundle\Messenger\DumpSitemapMessage;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+class DefaultController extends AbstractController
+{
+    public function index(MessageBusInterface $bus)
```
yann-eugone, line 33: Since the rework of our tests, we introduced an integration test suite (which is a minimal symfony project).
I'd like this snippet to be integrated in a test of that suite, could you write some test like that ?

Resources/config/services.xml
```diff
@@ -47,6 +47,13 @@
             <argument type="service" id="presta_sitemap.generator" />
             <argument>%presta_sitemap.timetolive%</argument>
         </service>
+
+        <service id="presta_sitemap.messenger.message_handler" class="Presta\SitemapBundle\Messenger\MessageHandler">
```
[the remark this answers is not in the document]
norkunas: Maybe we just could remove it in the extension if the messenger is not installed?

Messenger/DumpSitemapMessage.php
```diff
@@ -0,0 +1,100 @@
...
+
+    public function __construct(string $section = null, string $baseUrl = null, string $targetDir = null, array $options = [])
+    {
+        $this->section = $section;
+        $this->baseUrl = $baseUrl;
+        $this->targetDir = $targetDir;
+        $this->options = $options;
+    }
+
+    public function withSection(string $section = null): self
```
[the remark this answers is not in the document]
norkunas: :+1: just wanted to check what's your opinion will be on this :)

Messenger/MessageHandler.php
```diff
@@ -0,0 +1,94 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * Message handler to handle DumpSitemapMessage asynchronously or synchronously in background
+ *
+ * @author Tomas Norkūnas <norkunas.tom@gmail.com>
+ */
+class MessageHandler implements MessageHandlerInterface
```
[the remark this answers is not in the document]
norkunas: I've followed the Symfony mailer path and named it like this.

Tests/Unit/Messenger/DumpSitemapMessageTest.php
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+namespace Presta\SitemapBundle\Tests\Unit\Messenger;
+
+use PHPUnit\Framework\TestCase;
+use Presta\SitemapBundle\Messenger\DumpSitemapMessage;
+
+class DumpSitemapMessageTest extends TestCase
+{
+    public function testConstructWithProvidedData()
```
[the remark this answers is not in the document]
norkunas: Hmm, I'm not sure I understand what strict types you expect here?

Messenger/MessageHandler.php
```diff
@@ -0,0 +1,94 @@
...
+    private $defaultTarget;
+
+    public function __construct(RouterInterface $router, DumperInterface $dumper, string $defaultTarget)
+    {
+        $this->router = $router;
+        $this->dumper = $dumper;
+        $this->defaultTarget = $defaultTarget;
+    }
+
+    public function __invoke(DumpSitemapMessage $message)
```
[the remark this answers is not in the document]
norkunas: I'd prefer to leave refactoring to a follow-up PR :slightly_smiling_face:

norkunas:
And for the tests - what path should we follow? Add a flag on travis to remove the messenger dependency on the unsupported Symfony versions?

Resources/config/services.xml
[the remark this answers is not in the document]
norkunas: Done it in the way you wanted :)

Resources/doc/7-messenger-integration.md
norkunas: I've added some integration tests :)

norkunas:
I've managed to make build green, except that travis is red because of deprecations.

Messenger/MessageHandler.php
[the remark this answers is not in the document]
yann-eugone: For Symfony mailer, it is pretty obvious what MessageHandler is doing, here... Not that sure.

Messenger/MessageHandler.php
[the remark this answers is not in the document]
yann-eugone: Understood

Tests/Integration/config/messenger.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+    messenger:
+        transports:
+            async: 'in-memory://'
+        routing:
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage': async
+
+services:
```
yann-eugone, line 8: Is this required ?
Maybe Im' missing something, but the service should have been registered by the bundle.

Tests/Unit/Messenger/DumpSitemapMessageTest.php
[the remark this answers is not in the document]
yann-eugone: `: void` for tests return types

Tests/Integration/tests/MessengerTest.php
```diff
@@ -0,0 +1,169 @@
...
+
+        $blog = $this->section('blog', $gzip);
+        self::assertFileNotExists($blog, 'Sitemap "blog" section file does not exists before dump');
+
+        $archives = $this->section('archives', $gzip);
+        $archives0 = $this->section('archives_0', $gzip);
+        self::assertFileNotExists($archives, 'Sitemap "archive" section file does not exists before dump');
+        self::assertFileNotExists($archives0, 'Sitemap "archive_0" section file does not exists before dump');
+
+        /** @var MessageBusInterface $messageBus */
```
yann-eugone, line 96: This is not the kind of thing I was thinking about.

My thought was more like : adding a controller/command in the integration test project that will actually dispatch the message.
That controller/command should be called using same ways as done in Dump & Http tests

norkunas: Hey, I want to double check this with you. For me the way you described is just only that I'd move the dispatching elsewhere but everything would still be the same, so I don't understand why you want this :)

yann-eugone: The integration suite we introduced aims to test the code "like a developer using the bundle will".

The suite rely on actual/minimal symfony project. We have actual controllers, actual listener...
I'd like your test to be the same : actual controller, dispatching message on actual (synchronous) bus.

The test should call the controller using browserkit, and assert sitemap files are created and filled as expected.
It should be a mix of both the [Http](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Tests/Integration/tests/HttpTest.php) for calling your controller & [Cli](https://github.com/prestaconcept/PrestaSitemapBundle/blob/master/Tests/Integration/tests/CliTest.php) for assertion.

Tell me if this is clear enough ?
(writing on mobile is hard, sorry if you have difficulties to read me)

DependencyInjection/PrestaSitemapExtension.php
```diff
@@ -43,6 +44,10 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('route_annotation_listener.xml');
         }
 
+//        if (class_exists(MessageHandlerInterface::class)) {
+            $loader->load('messenger.xml');
+//        }
+
```
norkunas, line 50: This condition fails because when running phpunit it returns false. Maybe it's better to make a boolean option for enabling messenger integration then?

Tests/Integration/config/messenger.yaml
```diff
@@ -0,0 +1,12 @@
...
+        transports:
+            async: 'in-memory://'
+        routing:
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage': async
+#
+#services:
+#    Presta\SitemapBundle\Messenger\DumpSitemapMessageHandler:
+#        arguments:
+#            $defaultTarget: '%kernel.project_dir%/public'
+#        tags: ['messenger.message_handler']
```
norkunas, line 12: I remember now why I needed to add this. The messenger configuration for testing was imported based on message bus existence, so in phpunit it was returning false, removed the condition to check how it imports

Tests/Integration/tests/MessengerTest.php
norkunas: Is it the way now how you wanted? :)

norkunas: Is it now the way you wanted? :)

DependencyInjection/PrestaSitemapExtension.php
yann-eugone: Maybe it is just because you used `class_exists` instead of `interface_exists` ?

Tests/Integration/src/ContainerConfiguratorTrait.php
```diff
@@ -19,6 +20,10 @@ protected function configureContainer(ContainerConfigurator $container): void
             $container->import($confDir . '/{services}' . self::CONFIG_EXTS);
             $container->import($confDir . '/{services}_' . $this->environment . self::CONFIG_EXTS);
             $container->import($confDir . '/routing.yaml');
+
+            if (class_exists(MessageBusInterface::class)) {
```
yann-eugone, line 24: Same thing here : `class_exists` used on an interface instead of `interface_exists`

Tests/Integration/src/ContainerConfiguratorTrait.php
```diff
@@ -36,6 +41,10 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
             if (self::VERSION_ID >= 40200) {
                 $loader->load($confDir . '/routing.yaml');
             }
+
+            if (class_exists(MessageBusInterface::class)) {
```
yann-eugone, line 45: Same thing here : `class_exists` used on an interface instead of `interface_exists`

Tests/Integration/tests/MessengerTest.php
```diff
@@ -0,0 +1,138 @@
...
+use Symfony\Component\Messenger\Worker;
+
+class MessengerTest extends SitemapTestCase
+{
+    private const PUBLIC_DIR = __DIR__ . '/../public';
+    private const GET = Request::METHOD_GET;
+
+    protected function setUp(): void
+    {
+        if (!class_exists(MessageBusInterface::class)) {
```
yann-eugone, line 21: Same thing here : `class_exists` used on an interface instead of `interface_exists`

Tests/Integration/tests/MessengerTest.php
yann-eugone: Yes this is the way we designed that suite. :)
I believe that the worker part is somehow required, will try to find out an other way later on

Tests/Integration/config/messenger.yaml
```diff
@@ -0,0 +1,14 @@
...
+    messenger:
+        transports:
+            async: 'in-memory://'
+        routing:
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage': async
+
+services:
+    Presta\SitemapBundle\Tests\Integration\Controller\MessengerController:
+        tags: ['controller.service_arguments']
+#    Presta\SitemapBundle\Messenger\DumpSitemapMessageHandler:
```
yann-eugone, line 11: The previous comment was refering to this particular service, you may want to remove these lines now.

norkunas:
Updated.

norkunas:
More deprecations are coming from the symfony itself. So either we should ignore them or set `SYMFONY_DEPRECATIONS_HELPER="max[total]=11"`

yann-eugone:
Github actions was added to the bundle as a replacement to Travis, but we keep both for a period.

For what I've seen, these deprecations are not direct.
Will fix this in an other PR (will end up with some `SYMFONY_DEPRECATIONS_HELPER` env value).

Thank you for this good job here ! :)

PR #242, opened 2020-08-11
Update tests CS

norkunas:
- Add return `void` where applicable
- Remove unused imports
- Use `self::` instead of `$this->` where applicable for consistency
` Use `$this->expectException` instead of deprecated `@expectedException` annotation
- Remove useless phpdocs where argument typehints already declared


yann-eugone:
@norkunas you are on :fire: :+1:

PR #250, opened 2020-08-19
Added Dumper service unit tests

yann-eugone:
Unit tests for `\Presta\SitemapBundle\Service\Dumper` were missing.
This PR fixes this.

### DIFF
--- a/Tests/Unit/Service/DumperTest.php
+++ b/Tests/Unit/Service/DumperTest.php
@@ -1,0 +1,265 @@
+<?php
+
+namespace Presta\SitemapBundle\Tests\Unit\Service;
+
+use PHPUnit\Framework\TestCase;
+use Presta\SitemapBundle\Event\SitemapPopulateEvent;
+use Presta\SitemapBundle\Service\Dumper;
+use Presta\SitemapBundle\Sitemap\Url\UrlConcrete;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\Filesystem\Filesystem;
+
+class DumperTest extends TestCase
+{
+    private const DUMP_DIR = __DIR__ . '/.artifacts';
+
+    /**
+     * @var EventDispatcher
+     */
+    private $eventDispatcher;
+
+    /**
+     * @var Filesystem
+     */
+    private $filesystem;
+
+    /**
+     * @var Dumper
+     */
+    private $dumper;
+
+    private static function createDir(): void
+    {
+        if (!\is_dir(self::DUMP_DIR) && !@\mkdir(self::DUMP_DIR)) {
+            throw new \LogicException(\sprintf('Cannot create %s dir', self::DUMP_DIR));
+        }
+    }
+
+    private static function removeDir(): void
+    {
+        if (!\is_dir(self::DUMP_DIR)) {
+            return;
+        }
+
+        foreach (\scandir(self::DUMP_DIR) as $file) {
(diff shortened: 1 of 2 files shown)


Tests/Unit/Service/DumperTest.php
```diff
@@ -0,0 +1,265 @@
...
+     * @var Filesystem
+     */
+    private $filesystem;
+
+    /**
+     * @var Dumper
+     */
+    private $dumper;
+
+    private static function createDir(): void
```
J-Ben87, line 31: I guess you didn't take advantage of the Symfony `Filesystem` features here to reduce dependencies in the unit tests?

Tests/Unit/Service/DumperTest.php
```diff
@@ -0,0 +1,265 @@
...
+        <loc>https://acme.org/sitemap.default.xml.gz</loc>
+        <!-- missing <lastmod> tag -->
+    </sitemap>
+</sitemapindex>
+XML
+            ,
+        ];
+    }
+
+    private function assertGeneratedSitemap(
```
J-Ben87, line 187: You should make it `static`

Tests/Unit/Service/DumperTest.php
```diff
@@ -0,0 +1,265 @@
...
+                'Sitemap "blog" section file does not exists after dumping "default" section'
+            );
+            self::assertFileNotExists(
+                $blog0,
+                'Sitemap "blog_0 section file does not exists after dumping "default" section'
+            );
+        }
+    }
+
+    private function defaultListener(): \Closure
```
J-Ben87, line 234: You should make it `static`

Tests/Unit/Service/DumperTest.php
```diff
@@ -0,0 +1,265 @@
...
+                $urls->addUrl(new UrlConcrete('https://acme.org'), 'default');
+                $urls->addUrl(new UrlConcrete('https://acme.org/products'), 'default');
+                $urls->addUrl(new UrlConcrete('https://acme.org/contact'), 'default');
+                $urls->addUrl(new UrlConcrete('https://acme.org/team'), 'default');
+                $urls->addUrl(new UrlConcrete('https://acme.org/jobs'), 'default');
+            }
+        };
+    }
+
+    private function blogListener(): \Closure
```
J-Ben87, line 249: You should make it `static`

Tests/Unit/Service/DumperTest.php
J-Ben87, line 31: Shouldn't you move it after the `public` methods to comply with standards?

yann-eugone: You're right. I've no strong opinion on this particular point.
Maybe it's a bit overkill... What do you think?

yann-eugone: Not sure why I had this in my mind, but static was a whole section.
I was not able to find anything on this particular point, and after rereading the code this sounds weird.
Will fix

J-Ben87: Your code is quite simple and it's easy to understand.
It's obviously more robust to reduce the dependencies, especially when supporting several versions of Symfony.
One could assume that the component behavior may change from one version to another.

But because the Symfony `Filesystem` component is already a dependency of the project, because it's actually really stable and because I don't like to "reinvent the wheel", if I had been writing this test, I would have used it.

J-Ben87 (approved):

PR #251, opened 2020-08-20
Static routes translation alternate

yann-eugone:
As proposed in #205, I took this on my own to finish the job.

Idea is the following : 
- You configure the bundle to add alternate on routes with : 
  - allowed locales
  - default locale
  - technology used (symfony or jms are supported)
- Static routes listener (known as annotation listener) will trigger an event for each registered routes (can be used for something else) and an other listener will generate the multilang url if required


yann-eugone:
@fabianoroberto here is the version with tests.
It changes a bit since you wrote it, but url event was something I've always wanted, it helps decoupling code.

Please review this PR and tell me if it is still what you needed ?

fabianoroberto:
Looks good also your new part to handle URL adding with event listeners. I saw a PR to introduce symfony/messenger into the project, an alternative should be use it to generate new urls, but for now is a good improvement

J-Ben87 (approved):

J-Ben87 (approved):

PR #257, opened 2020-09-04
doc(Installation): Use .yaml extension to be understood by Symfony

qmachard:
Hi,

I create this very little PR to update "Installation" documentation. In Symfony 5, to makes routes ok, we need to use `.yaml` extension instead of `.yml`.

Thanks for this very practical project,
Bye. 👋


yann-eugone:
> Sometimes the difference between success and failure is a single `;` character
> **Someone, someday, somewhere**

Thank you for this :)

PR #265, opened 2020-12-16
Fix access to segmented sitemaps

Polylogic:


yann-eugone:
Hello @Polylogic ! Thank you for contributing.

We still have some trouble here : 
 
- I don't understand what is the problem you are trying to fix
- There is no description of your work
- The tests are all failing

Please take some time to answer and fix these points, otherwise I will have to close this PR

Polylogic:
Hi. I'm not sure about tests. This patch is pretty straightforward and only applies to the on-demand sitemap generation
When you request a sitemap segment, for example http://domain.tld/sitemap.entries_2.xml , the value passed to Generator::populate($section) is "entries_2". Which means we need to cut off the segment number "_2" in order to handle SitemapPopulateEvent($this, $section) event correctly. The regular expression is the same as the one used in the Dumper::dump().

yann-eugone:
@Polylogic seems that failing tests are due to incompatible version of the used tools, I opened #266 that should fix the issue. After the PR is merged, I'd appriate if you can rebase your branch.

I see the problem you had, but I would like to double check the behavior : can you add a test case in both `unit` & `integration` ?

yann-eugone:
Alternative PR was done, with tests

PR #271, opened 2021-02-25
Add default locale itself as hreflang too when using alternate option

### LINKED ISSUE #270: Add default locale as hreflang when using alternate option
This code should be removed:

https://github.com/prestaconcept/PrestaSitemapBundle/blob/63f54b809cc2f3cb82bf16cc866b9ea24421700b/EventListener/StaticRoutesAlternateEventListener.php#L69-L71

**Reason**

Both from [Google docs](https://developers.google.com/search/docs/advanced/crawling/localized-versions):

> Each <url> element must have a child element <xhtml:link rel="alternate" hreflang="supported_language-code"> that lists every alternate version of the page, including itself.

> Each language version must list itself as well as all other language versions.

I can open PR if that's okay.

taylankasap:
Closes #270

### DIFF
--- a/EventListener/StaticRoutesAlternateEventListener.php
+++ b/EventListener/StaticRoutesAlternateEventListener.php
@@ -66,10 +66,6 @@ final class StaticRoutesAlternateEventListener implements EventSubscriberInterfa
             )
         );
         foreach ($this->options['locales'] as $alternate) {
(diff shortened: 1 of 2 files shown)


yann-eugone:
Nicely done :)
Thank you

PR #276, opened 2021-03-26
Updated Bundle structure according to new best practices

yann-eugone:
* All sources were moved to `src/` directory
* Composer autoloading rule was changed from **PSR4** `"Presta\\SitemapBundle\\": ""` to `"Presta\\SitemapBundle\\": "src/"`
* `Tests/` directory was renamed to `tests/`
* Documentation was moved from `Resources/doc/` to `doc/`
* Services & routes were moved from `Resources/config/` to `config/`
* Routing import was changed from `"@PrestaSitemapBundle/Resources/config/routing.yml"` to `"@PrestaSitemapBundle/config/routing.yml"`

PR #279, opened 2021-05-24
Fix docs

pana1990:


norkunas:
There is no `Resources` folder with a new Symfony bundle structure :wink:

yann-eugone:
Yes, and this was an intended change : https://github.com/prestaconcept/PrestaSitemapBundle/pull/276/files#diff-1916af928b365eba7d6e08ded7efff71d43227dfb55e1ed60475f4d719cfe974

Maybe you followed the install procedure from the `3.x` branch (which is the default branch now), but with a `2.x` version installed ?

PR #280, opened 2021-05-26
Fix .yaml routes configuration 1-installation.md

Nykilor:
The import had a not valid key config value which would make the endpoints of this library not work.


norkunas:
That is incorrect. See #279

Nykilor:
Ah yes true. I look now at my `composer.json` and I'm at 2.x not 3.x. 

My bad closing this one. Sorry for this.

PR #282, opened 2021-06-25
Enhance 3.x testing

yann-eugone:
- Standardise sources
- Update testing matrix to support PHP 8.0 & Symfony 5.3


.github/workflows/tests.yml
```diff
@@ -16,12 +16,12 @@ jobs:
         include:
           - php-version: 7.1
             symfony-version: 4.4.*
-          - php-version: 7.4
+          - php-version: 8.0
```
J-Ben87, line 19: Why did you remove the support for php 7.4?

src/Event/SitemapAddUrlEvent.php
```diff
@@ -15,7 +15,11 @@
 use Symfony\Contracts\EventDispatcher\Event;
 
 /**
- * Event to allow generation of static routes sitemap urls.
+ * Event called whenever an static url is about to be added to sitemap.
```
J-Ben87:
```suggestion
 * Event called whenever a static url is about to be added to sitemap.
```

src/Event/SitemapPopulateEvent.php
```diff
@@ -15,9 +15,10 @@
 use Symfony\Contracts\EventDispatcher\Event;
 
 /**
- * Manage populate event
+ * Event called whenever a sitemap build is asked.
```
J-Ben87:
```suggestion
 * Event called whenever a sitemap build is requested.
```

src/EventListener/StaticRoutesAlternateEventListener.php
```diff
@@ -8,6 +17,11 @@
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
+/**
+ * Listen to "presta_sitemap.add_url" event.
+ * Decorate translatable Url with multi-lang alternates.
```
J-Ben87:
```suggestion
 * Decorate translatable Url with multi-lang alternatives.
```

src/Sitemap/Url/GoogleMultilangUrlDecorator.php
```diff
@@ -14,11 +14,9 @@
 use Presta\SitemapBundle\Sitemap\Utils;
 
 /**
- * Decorate w/ google alternate language url guidelines
- *
- * @see http://support.google.com/webmasters/bin/answer.py?hl=en&answer=2620865
+ * Url decorator to add alternates to a sitemap Url.
```
J-Ben87:
```suggestion
 * Url decorator to add alternatives to a sitemap Url.
```

tests/Integration/config/packages/5.3/framework.yaml
```diff
@@ -0,0 +1,17 @@
+framework:
+    secret: '%env(APP_SECRET)%'
+    #csrf_protection: true
```
J-Ben87: Is it necessary to keep this commented line?

tests/Integration/config/packages/5.3/framework.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+    # Enables session support. Note that the session will ONLY be started if you read or write from it.
+    # Remove or comment this section to explicitly disable session support.
+    session:
+        handler_id: null
+        cookie_secure: auto
+        cookie_samesite: lax
+        storage_factory_id: session.storage.factory.mock_file
+
+    #esi: true
```
J-Ben87: Is it necessary to keep these commented lines?

.github/workflows/tests.yml
yann-eugone: The bundle is still compatible with PHP 7.4, this is all about having a shorter testing matrix : only lowest and highest versions tested

Do you think it can be error prone ?

tests/Integration/config/packages/5.3/framework.yaml
yann-eugone: Not at all, came from the recipe I copied from Symfony's repository.

I will clean these

.github/workflows/tests.yml
J-Ben87: Not at all, I was just wondering why you were removing that version which I was considering as a major version.

However I like keeping only the lowest and highest versions. Or maybe you could keep the lowest and highest for each major? Like `7.1` and `7.4`, and `8.0`... until the next minor is released?

J-Ben87 (approved):

.github/workflows/tests.yml
yann-eugone: Actually, there is no need to keep `7.4` in testing matrix, the idea was to keep lowest and highest PHP version for each lowest and highest Symfony version

PR #281, opened 2021-06-08
Fix routes import path in docs

AntonOkulov:

PR #285, opened 2021-10-20
docs(📝): change routing import path

CharlyPoppins:


norkunas:
See #279 (ref: #280 #281)

yann-eugone:
@CharlyPoppins thank you for the PR, but I think @norkunas noticed the thing.

Is there a chance you were using a `2.x` version with a `3.x` install procedure ?

CharlyPoppins:
oh **** you're right. I get confused between projets...

PR #287, opened 2021-10-22
Make the url generator available on the events

### LINKED ISSUE #286: Make url generator available on `SitemapPopulateEvent`
**Description**  
A very common use case for using this event is to use the url generator to generate URLs. I think it would be a nice quality of life improvement to make the url generator available on the event itself. If of interest, I can make a PR.

**Example**  
```php
$event->getUrlGenerator()->generate(...);
```

kbond:
Fixes #286. Tests pass still so hopefully that means this is BC. Before I finalize, I'd like to ensure this direction is acceptable.

TODO:
- [x] Add tests
- [x] Trigger deprecation warnings
- [x] Document

### DIFF
--- a/src/Event/SitemapAddUrlEvent.php
+++ b/src/Event/SitemapAddUrlEvent.php
@@ -11,7 +11,9 @@
 
 namespace Presta\SitemapBundle\Event;
 
+use LogicException;
 use Presta\SitemapBundle\Sitemap\Url\Url;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Contracts\EventDispatcher\Event;
 
 /**
@@ -49,13 +51,20 @@ class SitemapAddUrlEvent extends Event
     private $options;
 
     /**
-     * @param string               $route
-     * @param array<string, mixed> $options
+     * @var UrlGeneratorInterface|null
      */
-    public function __construct(string $route, array $options)
+    protected $urlGenerator;
+
+    /**
+     * @param string                     $route
+     * @param array<string, mixed>       $options
+     * @param UrlGeneratorInterface|null $urlGenerator
+     */
+    public function __construct(string $route, array $options, UrlGeneratorInterface $urlGenerator = null)
     {
         $this->route = $route;
         $this->options = $options;
+        $this->urlGenerator = $urlGenerator;
     }
 
     /**
@@ -122,5 +131,14 @@ class SitemapAddUrlEvent extends Event
     public function getOptions(): array
     {
         return $this->options;
+    }
+
+    public function getUrlGenerator(): UrlGeneratorInterface
+    {
+        if (!$this->urlGenerator) {
+            throw new LogicException('UrlGenerator was not set.');
+        }
+
+        return $this->urlGenerator;
     }
 }
--- a/src/Event/SitemapPopulateEvent.php
+++ b/src/Event/SitemapPopulateEvent.php
@@ -11,7 +11,9 @@
 
 namespace Presta\SitemapBundle\Event;
 
+use LogicException;
 use Presta\SitemapBundle\Service\UrlContainerInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Contracts\EventDispatcher\Event;
 
 /**
@@ -39,13 +41,23 @@ class SitemapPopulateEvent extends Event
     protected $section;
 
     /**
-     * @param UrlContainerInterface $urlContainer
-     * @param string|null           $section
+     * @var UrlGeneratorInterface|null
      */
-    public function __construct(UrlContainerInterface $urlContainer, string $section = null)
-    {
+    protected $urlGenerator;
+
+    /**
+     * @param UrlContainerInterface      $urlContainer
+     * @param string|null                $section
+     * @param UrlGeneratorInterface|null $urlGenerator
+     */
+    public function __construct(
+        UrlContainerInterface $urlContainer,
+        string $section = null,
+        UrlGeneratorInterface $urlGenerator = null
+    ) {
         $this->urlContainer = $urlContainer;
         $this->section = $section;
+        $this->urlGenerator = $urlGenerator;
     }
 
     /**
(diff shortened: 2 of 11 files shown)


src/Event/SitemapPopulateEvent.php
```diff
@@ -39,13 +40,20 @@ class SitemapPopulateEvent extends Event
...
      */
-    public function __construct(UrlContainerInterface $urlContainer, string $section = null)
+    protected $urlGenerator;
+
+    /**
+     * @param UrlContainerInterface      $urlContainer
+     * @param string|null                $section
+     * @param UrlGeneratorInterface|null $urlGenerator
+     */
+    public function __construct(UrlContainerInterface $urlContainer, string $section = null, UrlGeneratorInterface $urlGenerator = null)
```
kbond: Thinking we could trigger a deprecation warning if `$urlGenerator` is null - WDYT?

src/Event/SitemapAddUrlEvent.php
```diff
@@ -49,13 +50,20 @@ class SitemapAddUrlEvent extends Event
...
      */
-    public function __construct(string $route, array $options)
+    protected $urlGenerator;
+
+    /**
+     * @param string                     $route
+     * @param array<string, mixed>       $options
+     * @param UrlGeneratorInterface|null $urlGenerator
+     */
+    public function __construct(string $route, array $options, UrlGeneratorInterface $urlGenerator = null)
```
kbond, line 63: I assumed you would want this even to include the `$urlGenerator` as well. If so, we should trigger a deprecation warning here when it is null.

src/Service/AbstractGenerator.php
```diff
@@ -45,20 +46,27 @@ abstract class AbstractGenerator implements UrlContainerInterface
...
 
     /**
-     * @param EventDispatcherInterface $dispatcher
-     * @param int|null                 $itemsBySet
+     * @param EventDispatcherInterface   $dispatcher
+     * @param int|null                   $itemsBySet
+     * @param UrlGeneratorInterface|null $urlGenerator
      */
-    public function __construct(EventDispatcherInterface $dispatcher, int $itemsBySet = null)
+    public function __construct(EventDispatcherInterface $dispatcher, int $itemsBySet = null, UrlGeneratorInterface $urlGenerator = null)
```
kbond: Trigger deprecation when `null` here?

src/Event/SitemapAddUrlEvent.php
yann-eugone: Not sure what would be the use case here, but why not.

But yes I think a deprecation should be triggered here if `$urlGenerator` is null.

src/Event/SitemapAddUrlEvent.php
```diff
@@ -123,4 +131,9 @@ public function getOptions(): array
     {
         return $this->options;
     }
+
+    public function getUrlGenerator(): ?UrlGeneratorInterface
```
yann-eugone: IMO we should have a strict return type, and throw an exception if property is null.

Something that tell the developer what he did wrong.

src/Event/SitemapPopulateEvent.php
yann-eugone: Yes, 100% agreed

src/Event/SitemapPopulateEvent.php
```diff
@@ -65,4 +73,9 @@ public function getSection(): ?string
     {
         return $this->section;
     }
+
+    public function getUrlGenerator(): ?UrlGeneratorInterface
```
yann-eugone: Same has above: strict return type, and throw an exception if property is null.

src/Service/AbstractGenerator.php
yann-eugone: Yes, here too

yann-eugone (changes requested):
Sorry I'm late (long weekend :))

I think this is a very good start, thank you.

Do you planned to update/add tests in the same PR ?

src/Event/SitemapAddUrlEvent.php
kbond: Perfect, I thought the same thing.

kbond:
> Do you planned to update/add tests in the same PR ?

Yes, that was my plan.

kbond:
Ok, addressed the comments. Let me know if you'd like me to change the deprecation/exception messages.

For the tests, I was thinking the following:
1. Mock/inject the `UrlGeneratorInterface` in the unit tests to remove the self-deprecations
2. Adjust the integration test's `SitemapListener` to use the url generator from the event (instead of injecting)

Let me know if you'd like anything else.

src/Event/SitemapAddUrlEvent.php
yann-eugone: Night passed, I'm wondering whether these deprecations at the event level are making sense.

We agree that this is a DX thing ?
A developer that rely on an event the bundle produces will be able to use this feature.
But if he do not want to use this feature and produces is own event, we don't care, do we ?

composer.json
```diff
@@ -17,7 +17,8 @@
         "php": ">=7.1.3",
         "ext-simplexml": "*",
         "symfony/console": "^4.4|^5.0",
-        "symfony/framework-bundle": "^4.4|^5.0"
+        "symfony/framework-bundle": "^4.4|^5.0",
+        "symfony/deprecation-contracts": "^2.4"
```
yann-eugone: I'm not a big fan of adding a dependency in a minor version.
We usually have issues with dependencies version compatibility : X requires Y 1.x, but Z requires Y 2.x

So far we did not used this library, we were triggering deprecations manually.
Do you think you can go back to that strategy ?

yann-eugone:
>     1. Mock/inject the `UrlGeneratorInterface` in the unit tests to remove the self-deprecations

I know that this bundle has some bad practices in tests, but I'm trying to fix this.
I'm trying to apply the "don't mock what you don't own" approach.
I suggest you inject an implementation of the UrlGeneratorInterface, some actual object from symfony ecosystem, or a test dummy you will create.

>     2. Adjust the integration test's `SitemapListener` to use the url generator from the event (instead of injecting)

:+1:, 100% agree with this

yann-eugone (changes requested):

src/Event/SitemapAddUrlEvent.php
kbond: Sure, that's fair. I'll remove.

composer.json
kbond: Sure, I can do this.

kbond:
Ok, I believe this is ready for a final review. I believe the phpstan errors are unrelated.

yann-eugone (approved):
Thank you for this work !

PR #291, opened 2021-12-15
Support for Symfony 5.4 & 6.0

yann-eugone:
Following #289, this PR introduces support for Symfony 6 & rewrite testing matrix for newer versions

PR #292, opened 2021-12-16
Allow Symfony 6.x in composer.json

yann-eugone:
Missing part of the #291


yann-eugone:
@norkunas there it is

PR #295, opened 2022-01-21
Allow subscribing events by class

norkunas:
Subscribing by event names was and old way before subscribing by event class name was not supported, so this enables it.

### DIFF
--- a/src/DependencyInjection/PrestaSitemapExtension.php
+++ b/src/DependencyInjection/PrestaSitemapExtension.php
@@ -11,6 +11,8 @@
 
 namespace Presta\SitemapBundle\DependencyInjection;
 
+use Presta\SitemapBundle\Event\SitemapAddUrlEvent;
+use Presta\SitemapBundle\Event\SitemapPopulateEvent;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader;
--- a/doc/4-dynamic-routes-usage.md
+++ b/doc/4-dynamic-routes-usage.md
@@ -47,7 +47,7 @@ class SitemapSubscriber implements EventSubscriberInterface
     public static function getSubscribedEvents()
     {
         return [
-            SitemapPopulateEvent::ON_SITEMAP_POPULATE => 'populate',
+            SitemapPopulateEvent::class => 'populate',
         ];
     }
 
--- a/src/Event/SitemapAddUrlEvent.php
+++ b/src/Event/SitemapAddUrlEvent.php
@@ -27,6 +27,7 @@ class SitemapAddUrlEvent extends Event
 {
     /**
      * @Event("Presta\SitemapBundle\Event\SitemapAddUrlEvent")
+     * @deprecated since presta/sitemap-bundle 3.3, use `SitemapAddUrlEvent::class` instead.
      */
     public const NAME = 'presta_sitemap.add_url';
 
--- a/src/Event/SitemapPopulateEvent.php
+++ b/src/Event/SitemapPopulateEvent.php
@@ -26,6 +26,7 @@ class SitemapPopulateEvent extends Event
 {
     /**
      * @Event("Presta\SitemapBundle\Event\SitemapPopulateEvent")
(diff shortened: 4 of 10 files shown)


src/DependencyInjection/PrestaSitemapExtension.php
```diff
@@ -40,6 +42,19 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter($this->getAlias() . '.defaults', $config['defaults']);
         $container->setParameter($this->getAlias() . '.default_section', (string)$config['default_section']);
 
+        if ($container->hasParameter('event_dispatcher.event_aliases')) {
```
yann-eugone: Maybe I'm missing something, but I was pretty sure that we should never rely on others logic at extension time and use compiler pass instead...
Or maybe you just don't care here because there is a if, and every code in all the Symfony ecosystem should perform this array_merge trick (even Symfony) ?

yann-eugone (commented):
Thank you for this Tomas.

I'm 100% aware that event names is not "the way to go" since few Symfony version.
But I've been struggling with BC to be able to remove it from the source code.
Now that you added the ability to register listener for both event name and event class, how can we inform developer that rely on event name that they must update their code for next major ?

src/DependencyInjection/PrestaSitemapExtension.php
norkunas: I'm exactly not sure where this part should be. I just found out today that this parameter can be set for this case and I though that at last I could get rid of using event name with this bundle :D

If compiler pass is the better place I can add it there. In Symfony source code I didn't found any usages, so this is added so external bundles could add mappings.

norkunas:
> Thank you for this Tomas.
> 
> I'm 100% aware that event names is not "the way to go" since few Symfony version. But I've been struggling with BC to be able to remove it from the source code. Now that you added the ability to register listener for both event name and event class, how can we inform developer that rely on event name that they must update their code for next major ?

I think we could just add @deprecated on constants then? For example https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpKernel/HttpKernelInterface.php `MASTER_REQUEST` constant

src/DependencyInjection/Compiler/EventAliasMappingPass.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Presta\SitemapBundle\Event\SitemapAddUrlEvent;
+use Presta\SitemapBundle\Event\SitemapPopulateEvent;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Maps event classes to event names
+ */
+final class EventAliasMappingPass implements CompilerPassInterface
```
yann-eugone: After a quick search over Symfony's source, I've found that they already wrote a compiler pass for this : [`AddEventAliasesPass`](https://github.com/symfony/event-dispatcher/blob/5.4/DependencyInjection/AddEventAliasesPass.php)

I think we should use it instead of writing our own

tests/Integration/src/Listener/SitemapBlogListener.php
```diff
@@ -20,9 +20,8 @@
 use Presta\SitemapBundle\Sitemap\Url\UrlConcrete;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Symfony\Component\Routing\RouterInterface;
 
-final class SitemapListener implements EventSubscriberInterface
+final class SitemapBlogListener implements EventSubscriberInterface
```
yann-eugone: Now that we deprecated the constants, does this split still make sense ?
Because we are using the deprecated constant here, and I'm not sure this is a good idea.

yann-eugone (changes requested):

src/DependencyInjection/Compiler/EventAliasMappingPass.php
norkunas: So revert to previous way?
Edit: no - find a documented way in https://symfony.com/doc/current/event_dispatcher.html#event-aliases

yann-eugone (approved):

yann-eugone:
Thank you very much Tomas :)

PR #298, opened 2022-06-07
Fix Symfony 6.1 deprecation

norkunas:


tests/Integration/config/5.4/messenger.yaml
```diff
@@ -1,5 +1,6 @@
 framework:
     messenger:
+        reset_on_message: true
```
norkunas, line 3: Also to fix in tests:
```
Remaining indirect deprecation notices (1)
  1x: Since symfony/framework-bundle 5.4: Not setting the "framework.messenger.reset_on_message" configuration option is deprecated, it will default to "true" in version 6.0.
    1x in CliTest::testDumpSitemapUsingCLI from Presta\SitemapBundle\Tests\Integration\Tests
```

.github/workflows/tests.yml
```diff
@@ -26,6 +26,8 @@ jobs:
             symfony-version: 6.0.*
           - php-version: 8.1
             symfony-version: 6.0.*
```
yann-eugone: I'm not sure I want to run tests on 6.0 if we are running 6.1
In fact, all 6.0 jobs in this file should be moved on 6.1

tests/Integration/config/5.4/messenger.yaml
yann-eugone: Nice catch

yann-eugone (changes requested):
Thanks for this
Please also see the error spotted by PhpStan with `Finder::create` in `src/Service/Dumper.php`

.github/workflows/tests.yml
norkunas: OK

.github/workflows/tests.yml
```diff
@@ -45,7 +43,7 @@ jobs:
             symfony/http-kernel:${{ matrix.symfony-version }} \
             symfony/routing:${{ matrix.symfony-version }} \
             symfony/messenger:${{ matrix.symfony-version }}
-          composer update --no-interaction --no-progress --no-suggest
```
norkunas, line 48: > You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.

yann-eugone: Yes, it's kind of the thing I'm fixing everywhere these days... If you can, remove it !

norkunas:
I think it's good now, and phpstan errors should be fixed in another PR or moved to baseline

norkunas:
Also cannot use latest phpstan on 7.1 because it requires at least php 7.2

yann-eugone:
Fine, will fix it myself
Thanks

yann-eugone:
Released as of https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v3.3.1

PR #303, opened 2023-01-09
Fix Symfony 6.2 deprecation

franmomu:
When using the bundle with Symfony 6.2 there is this deprecation warning:

`
The "Presta\SitemapBundle\Messenger\DumpSitemapMessageHandler" class implements "Symfony\Component\Messenger\Handler\MessageHandlerInterface" that is deprecated since Symfony 6.2, use the {@see AsMessageHandler} attribute instead.
`

AFAIK the `MessageHandlerInterface` is just a marker to tell Symfony to autoconfigure the class, but since it's already defined and configured in 

https://github.com/prestaconcept/PrestaSitemapBundle/blob/61e6c07606e4a52dfd76a0b754b07e3cdd50e31a/config/messenger.xml#L7-L12

I think the interface is not necessary.

~Technically this is a BC break, but I guess this class is not meant to be extended by the user.~

Update: Following https://github.com/prestaconcept/PrestaSitemapBundle/pull/303#issuecomment-1384860212, it is resolved by adding `handles="Presta\SitemapBundle\Messenger\DumpSitemapMessage"`, there is no need to remove the interface for now.

### DIFF
--- a/config/messenger.xml
+++ b/config/messenger.xml
@@ -8,7 +8,7 @@
             <argument type="service" id="router" />
             <argument type="service" id="presta_sitemap.dumper" />
             <argument>%presta_sitemap.dump_directory%</argument>
-            <tag name="messenger.message_handler" />
+            <tag name="messenger.message_handler" handles="Presta\SitemapBundle\Messenger\DumpSitemapMessage" />
         </service>
     </services>
 
--- a/src/DependencyInjection/PrestaSitemapExtension.php
+++ b/src/DependencyInjection/PrestaSitemapExtension.php
@@ -17,7 +17,7 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
-use Symfony\Component\Messenger\Handler\MessageHandlerInterface;
+use Symfony\Component\Messenger\MessageBusInterface;
 
 /**
  * Load Bundle configuration, configure container parameters & services.
@@ -51,7 +51,7 @@ class PrestaSitemapExtension extends Extension
             }
         }
 
-        if (interface_exists(MessageHandlerInterface::class)) {
+        if (interface_exists(MessageBusInterface::class)) {
             $loader->load('messenger.xml');
         }
 
--- a/tests/Integration/config/6.2/framework.yaml
+++ b/tests/Integration/config/6.2/framework.yaml
@@ -1,0 +1,14 @@
+framework:
+    test: true
+    secret: '%env(APP_SECRET)%'
+    http_method_override: false
+    session:
+        handler_id: null
+        cookie_secure: auto
+        cookie_samesite: lax
+        storage_factory_id: session.storage.factory.mock_file
+    php_errors:
+        log: true
+    router:
+        utf8: true
+    cache: null
--- a/tests/Integration/config/6.2/messenger.yaml
+++ b/tests/Integration/config/6.2/messenger.yaml
@@ -1,0 +1,6 @@
+framework:
+    messenger:
+        transports:
+            async: 'in-memory://'
+        routing:
+            'Presta\SitemapBundle\Messenger\DumpSitemapMessage': async
--- a/tests/Integration/config/6.2/presta_sitemap.yaml
+++ b/tests/Integration/config/6.2/presta_sitemap.yaml
@@ -1,0 +1,9 @@
+presta_sitemap:
+    default_section: static
+    dump_directory: "%kernel.project_dir%/public"
+    items_by_set: 10
+    alternate:
+        enabled: true
+        default_locale: en
+        locales: [en, fr]
+        i18n: symfony
--- a/tests/Integration/config/6.2/routes/annotations.yaml
+++ b/tests/Integration/config/6.2/routes/annotations.yaml
@@ -1,0 +1,3 @@
+controllers:
+    resource: ../../../src/Controller/
+    type: annotation
--- a/tests/Integration/config/6.2/routes/presta_sitemap.yaml
+++ b/tests/Integration/config/6.2/routes/presta_sitemap.yaml
@@ -1,0 +1,2 @@
+presta_sitemap:
+    resource: "@PrestaSitemapBundle/config/routing.yml"
--- a/tests/Integration/config/6.2/routes/translated.yaml
+++ b/tests/Integration/config/6.2/routes/translated.yaml
@@ -1,0 +1,7 @@
+about:
+  path:
+    en: /about
+    fr: /a-propos
+  defaults: { _controller: \Presta\SitemapBundle\Tests\Integration\Controller\StaticController::about }
+  options:
+    sitemap: true
--- a/tests/Integration/config/6.2/routes/xml.xml
+++ b/tests/Integration/config/6.2/routes/xml.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <route id="xml" path="/company">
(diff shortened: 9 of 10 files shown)


norkunas:
Maybe we should add `handles="Presta\SitemapBundle\Messenger\DumpSitemapMessage"` as Symfony does in it's core configurations?

norkunas:
Edit: oh because ecd355fbfa0e2201ffa7625673ffcdffda336756 was not released yet..

Also I saw another two deprecation logs but they says the same:

> The "Symfony\Component\Console\Command\Command::$defaultName" property is considered final. You should not override it in "Presta\SitemapBundle\Command\DumpSitemapsCommand".

> Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Presta\SitemapBundle\Command\DumpSitemapsCommand" class instead.

franmomu:
> Maybe we should add `handles="Presta\SitemapBundle\Messenger\DumpSitemapMessage"` as Symfony does in it's core configurations?

Apparently it's not necessary: https://symfony.com/doc/current/messenger.html#manually-configuring-handlers

```xml
<service id="App\MessageHandler\SmsNotificationHandler">
     <!-- handles is only needed if it can't be guessed by type-hint -->
     <tag name="messenger.message_handler"
          handles="App\Message\SmsNotification"/>
</service>
```
but if they have it I don't know why 🤷‍♂️

norkunas:
When it is declared Symfony does not need to guess what this handler handles, so imho from perf side it would be better even it's unnoticeable

franmomu:
> When it is declared Symfony does not need to guess what this handler handles, so imho from perf side it would be better even it's unnoticeable

There is no need then to remove the interface 👍

maho125:
I got another deprecation with symfony v6.2.6:

`[2023-02-17T23:28:29.241934+00:00] deprecation.INFO: User Deprecated: Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Presta\SitemapBundle\Command\DumpSitemapsCommand" class instead. {"exception":"[object] (ErrorException(code: 0): User Deprecated: Since symfony/console 6.1: Relying on the static property \"$defaultName\" for setting a command name is deprecated. Add the \"Symfony\\Component\\Console\\Attribute\\AsCommand\" attribute to the \"Presta\\SitemapBundle\\Command\\DumpSitemapsCommand\" class instead. at /var/www/coupodo.com/html/vendor/symfony/console/Command/Command.php:85)"} []`

Here is fix for it:

`<?php

/*
 * This file is part of the PrestaSitemapBundle package.
 *
 * (c) PrestaConcept <https://prestaconcept.net>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace Presta\SitemapBundle\Command;

use Presta\SitemapBundle\Service\DumperInterface;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputArgument;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Input\InputOption;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\Routing\RouterInterface;
use Symfony\Component\Console\Attribute\AsCommand;

/**
 * Command to dump the sitemaps to provided directory
 */
#[AsCommand(name: 'presta:sitemaps:dump')]
class DumpSitemapsCommand extends Command
{
    /**
     * @var RouterInterface
     */
    private $router;

    /**
     * @var DumperInterface
     */
    private $dumper;

    /**
     * @var string
     */
    private $defaultTarget;

    public function __construct(RouterInterface $router, DumperInterface $dumper, string $defaultTarget)
    {
        $this->router = $router;
        $this->dumper = $dumper;
        $this->defaultTarget = $defaultTarget;

        parent::__construct();
    }

    /**
     * @inheritdoc
     */
    protected function configure(): void
    {
        $this
            ->setDescription('Dumps sitemaps to given location')
            ->addOption(
                'section',
                null,
                InputOption::VALUE_REQUIRED,
                'Name of sitemap section to dump, all sections are dumped by default'
            )
            ->addOption(
                'base-url',
                null,
                InputOption::VALUE_REQUIRED,
                'Base url to use for absolute urls. Good example - http://acme.com/, bad example - acme.com.' .
                ' Defaults to router.request_context.host parameter'
            )
            ->addOption(
                'gzip',
                null,
                InputOption::VALUE_NONE,
                'Gzip sitemap'
            )
            ->addArgument(
                'target',
                InputArgument::OPTIONAL,
                'Location where to dump sitemaps. Generated urls will not be related to this folder.',
                $this->defaultTarget
            );
    }

    /**
     * @inheritdoc
     */
    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        /** @var string $targetDir */
        $targetDir = $input->getArgument('target');
        $targetDir = rtrim($targetDir, '/');

        /** @var string|null $baseUrl */
        $baseUrl = $input->getOption('base-url');
        if ($baseUrl) {
            $baseUrl = rtrim($baseUrl, '/') . '/';

            //sanity check
            if (!parse_url($baseUrl, PHP_URL_HOST)) {
                throw new \InvalidArgumentException(
                    'Invalid base url. Use fully qualified base url, e.g. http://acme.com/',
                    -1
                );
            }

            // Set Router's host used for generating URLs from configuration param
            // There is no other way to manage domain in CLI
            $request = Request::create($baseUrl);
            $this->router->getContext()->fromRequest($request);
        } else {
            $baseUrl = $this->getBaseUrl();
        }

        /** @var string|null $section */
        $section = $input->getOption('section');
        if ($section) {
            $output->writeln(
                sprintf(
                    "Dumping sitemaps section <comment>%s</comment> into <comment>%s</comment> directory",
                    $section,
                    $targetDir
                )
            );
        } else {
            $output->writeln(
                sprintf(
                    "Dumping <comment>all sections</comment> of sitemaps into <comment>%s</comment> directory",
                    $targetDir
                )
            );
        }

        $options = [
            'gzip' => (bool)$input->getOption('gzip'),
        ];
        $filenames = $this->dumper->dump($targetDir, $baseUrl, $section, $options);

        if (!is_array($filenames)) {
            $output->writeln(
                "<error>No URLs were added to sitemap by EventListeners</error>" .
                " - this may happen when provided section is invalid"
            );

            return 1;
        }

        $output->writeln("<info>Created/Updated the following sitemap files:</info>");
        foreach ($filenames as $filename) {
            $output->writeln("    <comment>$filename</comment>");
        }

        return 0;
    }

    private function getBaseUrl(): string
    {
        $context = $this->router->getContext();

        if ('' === $host = $context->getHost()) {
            throw new \RuntimeException(
                'Router host must be configured to be able to dump the sitemap, please see documentation.'
            );
        }

        $scheme = $context->getScheme();
        $port = '';

        if ('http' === $scheme && 80 != $context->getHttpPort()) {
            $port = ':' . $context->getHttpPort();
        } elseif ('https' === $scheme && 443 != $context->getHttpsPort()) {
            $port = ':' . $context->getHttpsPort();
        }

        return rtrim($scheme . '://' . $host . $port, '/') . '/';
    }
}`

norkunas:
Could this get merged? :blush:

yann-eugone:
Will fix https://github.com/prestaconcept/PrestaSitemapBundle/issues/307

yann-eugone (approved):

yann-eugone:
I will take some time to see if something more is required for us to release it, but trying to do it ASAP

yann-eugone:
Released as of https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v3.3.1

norkunas:
The deprecation about `MessageHandlerInterface` is still triggered. I guess we should conditionally create the class based on  symfony version something like:
```php
if (Kernel::VERSION_ID >= 60200) {
    class DumpSitemapMessageHandler {}
} else {
    class DumpSitemapMessageHandler implements MessageHandlerInterface {}
}
```
to avoid triggering it. or check if we need to implement it at all.

yann-eugone:
Here we go again with these ugly solutions ^^
I'm wondering whether it is actually BC ?
If we remove the interface, but it is still working on EVERY symfony version we support, it should be OK right ?
This code was never intended to be used directly, or extended, or...
I believe that if someone has some static analysis tools plugged on, relying on this interface, he should be pissed we removed the interface, but does that exists anyway ?
Do we have examples on how the community has handled these changes ?

norkunas:
I leaning to think that `handles` part in config is enough and should work

PR #331, opened 2024-01-24
remove $ so gitclip works

tacman:


yann-eugone (approved):

yann-eugone:
I will think about it for all the packages I maintain, thanks

PR #337, opened 2024-02-08
Use https for real URLs

norkunas:


yann-eugone:
That is indeed a good idea, I want to merge it
I just wonder if there is any risk releasing it in a bugfix release, WDYT ?

yann-eugone (approved):

norkunas:
If people use full xml output fixtures with those links then it will break for them totally, I'm using like that by myself, but I don't mind to break them, it's easy to fix and it doesn't really break any functionality :)

yann-eugone:
OK, so at least publish it in a minor version

PR #344, opened 2024-06-11
Fix DI Extension class deprecation with Symfony 7.1

norkunas:
The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Presta\SitemapBundle\DependencyInjection\PrestaSitemapExtension".

### DIFF
--- a/src/DependencyInjection/PrestaSitemapExtension.php
+++ b/src/DependencyInjection/PrestaSitemapExtension.php
@@ -11,12 +11,10 @@
 
 namespace Presta\SitemapBundle\DependencyInjection;
 
-use Presta\SitemapBundle\Event\SitemapAddUrlEvent;
-use Presta\SitemapBundle\Event\SitemapPopulateEvent;
 use Symfony\Component\Config\FileLocator;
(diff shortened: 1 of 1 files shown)


yann-eugone:
I've fixed PhpStan in https://github.com/prestaconcept/PrestaSitemapBundle/pull/345

yann-eugone:
Found the PR where that class was deprecated https://github.com/symfony/symfony/pull/53801
Looks like a "side effect", but still need to change the parent

yann-eugone (approved):

yann-eugone:
Thank you @norkunas :pray:

mboron83:
Hi. @yann-eugone, tell me - when are You going to publish next stable release with this change.

yann-eugone:
Hello @mboron83 

Just did https://github.com/prestaconcept/PrestaSitemapBundle/releases/tag/v4.1.2

PR #354, opened 2025-07-31
Fix Route annotation/attribute reference in documentation

tacman:
fixed wrong class.


yann-eugone (approved):
Nice catch, thank you

PR #356, opened 2025-12-02
Allow Symfony 8

skmedix:
This PR is big one because it encompasses all the necessary changes that have been building up over time.

In Symfony 7.4, XML configurations are deprecated and removed in Symfony 8.0, so I moved the configurations to PHP, as this will not result in a BC. In Symfony 8.0, there are no more annotations, but since this 4.x bundle supports Symfony versions above 6.4, I changed the controllers to use attributes. I also changed routing.yml to use php. However, to avoid introducing a breaking change, I left both routing.yml and the new routing.yml, as both should work. Please review it thoroughly in case I have overlooked something.


.github/workflows/tests.yml
```diff
@@ -20,16 +20,16 @@ jobs:
...
-          - php-version: 8.4
-            symfony-version: 6.4.*
           - php-version: 8.2
-            symfony-version: 7.2.*
+            symfony-version: 7.4.*
+          - php-version: 8.4
+            symfony-version: 7.4.*
           - php-version: 8.4
-            symfony-version: 7.3.*
+            symfony-version: 8.0.*
```
yann-eugone, line 32: I know it's not written anywhere but our testing matrix policy is

every Symfony LTS version, + latest compatible version
each with lowest and hight compatible PHP version

I'd like to keep it that way

config/alternate_listener.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+use Presta\SitemapBundle\Event\SitemapAddUrlEvent;
+use Presta\SitemapBundle\EventListener\StaticRoutesAlternateEventListener;
+
+return static function (ContainerConfigurator $container) {
+    $services = $container->services();
+    $parameters = $container->parameters();
```
yann-eugone: please remove all the unused vars from the services files

tests/Integration/config/6.4/routes/yaml.yaml
```diff
@@ -4,3 +4,12 @@ yaml:
   options:
     sitemap:
       section: static
+
+company:
```
yann-eugone: Is this a route moved from `xml.xml` because XML routing was deprecated ?

tests/Integration/src/Controller/ArchivesController.php
```diff
@@ -12,13 +12,10 @@
 namespace Presta\SitemapBundle\Tests\Integration\Controller;
 
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Routing\Attribute\Route;
 
 final class ArchivesController
 {
-    /**
```
yann-eugone, line 19: maybe I got it wrong, but with Symfony 6.4, Annotation routing is still a thing ?
if so, we have to continue testing it

yann-eugone (changes requested):
Hi there
First of all, thank you for reaching out
It's always a pleasure to have contributors :+1: 

I've set a review, and will try my best to help you finish it

.github/workflows/tests.yml
skmedix: If I understand this correctly, this should be better:
| Symfony | PHP Lowest | PHP Highest |
|---|---|---|
| 6.4 LTS | 8.1 | 8.5 |
| 7.4 LTS | 8.2 | 8.5 |
| 8.0 | 8.4 | 8.5 |

tests/Integration/config/6.4/routes/yaml.yaml
skmedix: Yeah, for 6.4 and 7.4 it wasn't necessary to remove xml.xml, reverted

.github/workflows/tests.yml
skmedix: Update: I had to lower the highest of 6.4 to PHP 8.3, as it was failing the CI. However, I don't have time to figure it out today :)

yann-eugone (approved):
Thank you very much for you help
I will try to see why you had to lower the PHP version on my own

benjaminroche4:
Please merge it !

yann-eugone:
Nope sorry, still need to understand why @skmedix had to remove some versions from the testing matrix

skmedix:
@yann-eugone I figured out that it wasn't installing the correct version of Symfony. This has been fixed, and I have also updated the tests to use PHP 8.5. Everything should be green now :)

yann-eugone:
Because tests are OK, I will just merge it like this
But I think there things I need to try on own before going to a stable release

So for the moment, if you want to have this bundle on Symfony `8.0` you will need to require the `4.x` dev version

yann-eugone:
@skmedix the issue you encountered with the Symfony version was fixed using more composer require, see https://github.com/prestaconcept/PrestaSitemapBundle/pull/357/commits/91d2ea10616e9da520487de8f69307358b3d4529

the issue with routing/annotation was fixed using double alias in each controller, see https://github.com/prestaconcept/PrestaSitemapBundle/pull/357/commits/6364bf5225e27dab7f179dda6c6bb2ac494c21d8

tacman:
I've updated my branch of the demo to Symfony 8

https://github.com/tacman/presta-sitemap-test-project

The dynamic sitemap is no longer working, so I added instructions to dump it to the README.  I switched the database to sqlite so that this would work:

```bash 
git clone git@github.com:tacman/presta-sitemap-test-project.git presta && cd presta
composer install
bin/console doctrine:schema:update --force --complete
bin/console doctrine:fixtures:load --no-interaction
symfony server:start -d
symfony open:local
```

no dynamic sitemaps, but

```bash 
bin/console presta:sitemaps:dump
symfony open:local
```

works.

yann-eugone:
I don't understand the issue you are facing.
What is failing exactly?

I just tried the `4.x-dev@dev` along with standard Symfony `8.0` on a very simple project with entities in the sitemap, and it just works.

tacman:
image


No sitemaps in the index until

bin/console presta:sitemaps:dump

image

yann-eugone:
I don't get your problem
That's exactly what `presta:sitemaps:dump` was created for : it collect all the URLs of your sitemaps, and create the according files in the public directory

So before you call that command, there is no reason for these files to exists

tacman:
I was under the impression that these were available dynamically, but dumping them make them static and was recommended for production.  This follows the Symfony AssetMapper model.

But I must have misunderstood dynamic v static.

tacman:
> Access sitemap on the fly with a symfony controller or Dump sitemap to files for faster sitemap

I think the issue might be the dynamic list of sitemap Urlsets, v. the sitemaps themselves.  The behavior shows above is that the sets aren't listed until they're dumped.

yann-eugone:
How is that list built?
If it relies on the filesystem, then you have no other option than dumping the sitemap to have it listed.

I believe the "on the fly" sitemaps are the one accessed using the routing and the controller provided by the bundle.
The content will be built whenever you will access the /sitemap.xml, but don't expect any kind of listing.

If you have other questions, please open an issue.

PR #357, opened 2025-12-16
Fix deprecations after update

yann-eugone:
Following #356, I had to check remaining deprecations, but have a closer look to some changes I did not understand on the PR

I tried to fix most of the things I saw

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,16 +43,15 @@ jobs:
           coverage: none
           php-version: ${{ matrix.php-version }}
 
-      - name: "symfony/flex is required to install the correct symfony version"
-        run: |
-          composer global config --no-plugins allow-plugins.symfony/flex true
-          composer global require symfony/flex
-
-      - name: "Configure Symfony version for symfony/flex"
-        run: composer config extra.symfony.require "${{ matrix.symfony-version }}"
-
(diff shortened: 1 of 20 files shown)


skmedix (commented):
Using two import aliases is a clever solution. Coming from a Java, I assumed this was not possible, which is why I introduced a class_alias in the Kernel, especially since that class already contained many version-based tweaks before. I am also curious why using two aliases with the same name works now, whereas using the same class names did not before. It is possible that this was a different issue altogether, and we could switch back to using Route.

I also noticed that you reverted the use of symfony/flex, which was ensuring that all Symfony components stayed on the same version line. As shown in your [CI run](https://github.com/prestaconcept/PrestaSitemapBundle/actions/runs/20263023075/job/58179370675), Symfony versions are now mixed again, which can led to the unexpected incompatibilities as I encountered before.

yann-eugone:
Symfony flex is a bit overkill for what we need, adding all the used package to the targeted version is much more efficient

I'm not sure what was the issue you were facing with these classes aliases.
Actually, you can import as many classes with the same name as you want, provided you provide a unique alias for each.

adumont-prestaconcept (approved):

norkunas:
Could we get a release? :slightly_smiling_face:

PR #363, opened 2026-07-11
Reset generator state after fetch to prevent URL accumulation in persistent runtimes

### LINKED ISSUE #362: Generator accumulates URLs across requests in persistent runtimes (FrankenPHP worker mode), producing duplicated sitemap entries
### Bug: `Generator` accumulates URLs across requests in persistent runtimes (FrankenPHP worker mode / RoadRunner), producing duplicated sitemap entries

**Bundle version:** 4.3.0
**Symfony:** 7.x/8.x
**PHP:** 8.5
**Runtime:** FrankenPHP **worker mode** (also affects RoadRunner / Swoole — any long-running worker)

#### Summary

When the sitemaps are served live by `SitemapController` under a persistent runtime, every URL in a section gets duplicated **once per request handled by the same worker process**. After N requests, each `<loc>` appears N times.

Real example from a production `sitemap.artiste.xml`: 30 `<loc>` entries for only 6 unique URLs — each repeated exactly 5×.

#### Root cause

`Generator` is registered as a **shared** service, and it is **stateful**: `fetch()` calls `populate()`, which dispatches `SitemapPopulateEvent`, and listeners append their URLs into `AbstractGenerator::$urlsets`. However, `Generator::fetch()` never resets `$urlsets` / `$root` before populating.

In classic PHP-FPM this is harmless because the container is rebuilt on every request. But in a **persistent runtime** the kernel and its shared services survive across requests, so the previously-collected URLs are still in `$urlsets`. `getUrlset()` returns the existing (already-populated) urlset and the listeners append the same URLs again:

```php
// AbstractGenerator::getUrlset()
public function getUrlset(string $name): Urlset
{
    if (!isset($this->urlsets[$name])) {
        $this->urlsets[$name] = $this->newUrlset($name);
    }
    return $this->urlsets[$name]; // reused across requests -> URLs pile up
}
```

Note this only affects the **live** controller path. The `Dumper` path is fine because `Dumper::dump()` calls `cleanup()` (which resets `$urlsets` / `$root`) in a `finally` block.

It's also worth noting that Symfony's `FrankenPhpWorkerRunner` does **not** call `services_resetter` between requests (it only clones the kernel when `FRANKENPHP_RESET_KERNEL` is enabled), so tagging the service `kernel.reset` is not a reliable fix on the user side either.

#### Steps to reproduce

1. Serve a Symfony app with PrestaSitemapBundle under FrankenPHP worker mode (`worker { file ./public/index.php }`), with at least one `SitemapPopulateEvent` listener adding URLs to a section.
2. Request the same section repeatedly, e.g.:
   ```
   for i in $(seq 1 6); do curl -s https://app.example/sitemap.mysection.xml | grep -c '<loc>'; done
   ```
3. The `<loc>` count grows (e.g. `307, 307, 307, 614, 614, 921`) while the number of *unique* URLs stays constant.

#### Suggested fix

Reset the generator's accumulated state at the start of `Generator::fetch()` (mirroring what `Dumper::dump()` already does via `cleanup()`), so each `fetch()` starts from a clean slate:

```php
public function fetch(string $name): ?XmlConstraint
{
    $this->urlsets = [];
    $this->root = null;

    // ... existing logic
}
```

Alternatively, make `Generator` implement `Symfony\Contracts\Service\ResetInterface` and tag `presta_sitemap.generator_default` with `kernel.reset` — though, as noted above, that tag isn't honored by the FrankenPHP worker runner, so resetting inside `fetch()` is the more robust option.

#### Workaround

For anyone hitting this before a fix lands: override the `presta_sitemap.generator.class` parameter with a subclass that clears the state in `fetch()`:

```php
final class StatelessSitemapGenerator extends Generator
{
    public function fetch(string $name): ?XmlConstraint
    {
        $this->urlsets = [];
        $this->root = null;

        return parent::fetch($name);
    }
}
```

```yaml
parameters:
    presta_sitemap.generator.class: App\Sitemap\StatelessSitemapGenerator
```

ker0x:
Fixes #362

## Problem

`Generator` is registered as a **shared** service and is **stateful**: `fetch()` calls `populate()`, which dispatches `SitemapPopulateEvent`, and listeners append their URLs into `AbstractGenerator::$urlsets`. However, `fetch()` never resets `$urlsets` / `$root` before returning.

Under classic PHP-FPM this is harmless because the container is rebuilt on every request. But in a **persistent runtime** (FrankenPHP worker mode, RoadRunner, Swoole) the kernel and its shared services survive across requests, so previously-collected URLs remain in `$urlsets` and listeners append the same URLs again — each `<loc>` ends up duplicated once per request handled by the same worker process.

Note this only affects the **live** controller path; the `Dumper` path is fine because `Dumper::dump()` already resets state via `cleanup()` in a `finally` block.

## Fix

Reset `$root` / `$urlsets` in a `finally` block once the section has been built, mirroring `Dumper::dump()`.

Resetting at the **end** of `fetch()` (rather than the start, as suggested in the issue) is deliberate: it preserves the public `addUrl()`-then-`fetch()` pattern within a single request — resetting at the start would discard URLs added manually before `fetch()` (a regression caught by `GoogleNewsUrlDecoratorTest`). The returned `Urlset` / `Sitemapindex` object is already built and held by the caller before the array is cleared, so returning it after cleanup is safe. The controller only performs a single `fetch()` per request.

## Test

Added `testFetchDoesNotAccumulateAcrossCalls`, which reuses one `Generator` instance across two `fetch()` calls (emulating a persistent worker) and asserts the URL count stays constant. Verified it fails without the fix and passes with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /.idea/
 /.phpcs-cache
+/.phpunit.cache
 /.phpunit.result.cache
 /composer.lock
 /vendor/
--- a/config/services.php
+++ b/config/services.php
@@ -21,7 +21,8 @@ return static function (ContainerConfigurator $container) {
             service('router'),
             '%presta_sitemap.items_by_set%',
         ])
-        ->call('setDefaults', ['%presta_sitemap.defaults%']);
+        ->call('setDefaults', ['%presta_sitemap.defaults%'])
+        ->tag('kernel.reset', ['method' => 'reset']);
 
     $services->set('presta_sitemap.dumper_default', '%presta_sitemap.dumper.class%')
         ->args([
--- a/src/Service/Generator.php
+++ b/src/Service/Generator.php
@@ -46,20 +46,25 @@ class Generator extends AbstractGenerator implements GeneratorInterface
      */
     public function fetch(string $name): ?XmlConstraint
     {
-        if ('root' === $name) {
-            $this->populate();
+        try {
+            if ('root' === $name) {
+                $this->populate();
 
-            return $this->getRoot();
+                return $this->getRoot();
+            }
+
+            $baseName = preg_replace('/(.*?)(_\d+)?/', '\1', $name);
+            $this->populate($baseName);
+
+            if (array_key_exists($name, $this->urlsets)) {
+                return $this->urlsets[$name];
+            }
+
+            return null;
+        } finally {
+            $this->root = null;
+            $this->urlsets = [];
         }
-
-        $baseName = preg_replace('/(.*?)(_\d+)?/', '\1', $name);
-        $this->populate($baseName);
-
-        if (array_key_exists($name, $this->urlsets)) {
-            return $this->urlsets[$name];
-        }
-
-        return null;
     }
 
     /**
--- a/src/Service/Generator.php
+++ b/src/Service/Generator.php
@@ -15,11 +15,12 @@ use Presta\SitemapBundle\Sitemap\Urlset;
 use Presta\SitemapBundle\Sitemap\XmlConstraint;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * Sitemap generator.
  */
-class Generator extends AbstractGenerator implements GeneratorInterface
+class Generator extends AbstractGenerator implements GeneratorInterface, ResetInterface
 {
     /**
      * @var UrlGeneratorInterface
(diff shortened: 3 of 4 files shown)


yann-eugone:
Should we use [`Symfony\Contracts\Service\ResetInterface`](https://github.com/symfony/contracts/blob/main/Service/ResetInterface.php) instead?

ker0x:
Added a second commit implementing `ResetInterface` as a complement to the in-`fetch()` reset:

- `Generator` now implements `Symfony\Contracts\Service\ResetInterface` (`reset()` clears `$root`/`$urlsets`), and `fetch()`'s `finally` delegates to it.
- The `presta_sitemap.generator_default` service is tagged `kernel.reset` (`method: reset`), so runtimes that invoke the `services_resetter` free the accumulated urlsets at the request boundary.

This is deliberately **belt-and-suspenders**, not the primary fix: some runners — notably FrankenPHP worker mode — don't call the resetter between requests, so a `ResetInterface`-only approach would leave the reported runtime broken. The reset performed **inside `fetch()`** remains the runtime-agnostic guarantee; `ResetInterface` adds the idiomatic hook and a memory benefit where it's honored.

Note: `symfony/service-contracts` (which provides `ResetInterface`) is pulled transitively via `symfony/framework-bundle` and isn't declared in `composer.json` — consistent with how the bundle already consumes `event-dispatcher`/`routing` symbols. Happy to declare it explicitly if you'd prefer.

